PR #11, opened 2026-06-16
map: fix server + SD tile rendering, add fetch diagnostics

Vybo:
**Big disclaimer**: Only SD card under M5Launcher was tested. I haven't tested LittleFS build for now. This is the reason for the draft state.

Fixes the Map tab showing no tiles in **either** source mode (tile server or microSD), and adds on-screen fetch diagnostics — serial is unreadable on this build (the companion binary protocol owns the UART), so the Map tab is the only window into fetch health.

**Root cause (server mode)**
`tilesFsLowSpace()` always measured the LittleFS `tiles` partition. When that partition doesn't mount and the cache falls back to the SD card, the partition object is unmounted → `totalBytes()` reads 0 → the low-space gate wedges on → every tile download is dropped before the HTTP GET (on-screen diag read `ok 0 / http 0 / wr S`). The guard now applies only to the LittleFS backend.

**Changes**
- **Low-space gate:** guard the active backend only; the SD-backed cache (gigabytes free) is never treated as full, and a 0-byte/unmounted FS no longer reads as "full".
- **Write verification:** the fetch worker now checks `f.write()`'s return — success was judged by bytes read from the network, so a short/failed SD write was counted as cached when the file was empty/truncated.
- **Read cap:** loader cap raised 80 → 100 KB to match the fetcher's write cap; dense tiles were saved then silently rejected on read.
- **SD PNG packs:** support `/tiles/<z>/<x>/<y>.png(.PNG)`, not just `/maps/osm`; mirrored in the zoom-availability probe.
- **Diagnostics:** the Map-tab overlay now shows the active backend, HTTP code, last write outcome, and open-fail / short-write counters.

Verified the T-Deck build (`LilyGo_TDeck_companion_radio_touch`); the SD-specific paths are `HAS_TDECK_GT911`-gated, so the Heltec V4 is unaffected. 

Please let me know if the Heltec gate should be removed, but I don't have access to that device.

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -12973,7 +12973,16 @@ static inline void tileCacheMkdir(const char* rel) {
 // core-0 fetch task, never the UI thread. Reads keep working, so already-cached
 // areas still display — only NEW tiles stop being cached.
 static bool tilesFsLowSpace() {
-  if (!s_tiles_fs_ready) return false;            // no LittleFS cache partition -> nothing to guard
+  if (!s_tiles_fs_ready || !s_tile_fs) return false;   // no cache backend -> nothing to guard
+  // This guard exists ONLY for the small (4.75 MB) LittleFS "tiles" partition,
+  // where a full FS faults inside lfs_alloc during the dir mkdir. It must inspect
+  // the ACTIVE backend, not s_tiles_fs unconditionally. On the SD fallback
+  // (s_tile_fs == &SD, when the tiles partition didn't mount) the LittleFS object
+  // is UNMOUNTED, so s_tiles_fs.totalBytes() reads 0 -> "0 free < 320 KB" -> the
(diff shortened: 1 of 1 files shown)


B0N3-GD:
Merged — thanks @Vybo, great catch and a genuinely thorough fix. 🙌 You'd independently nailed the same `tilesFsLowSpace()` root cause we were chasing, and the write-verification + read/write cap match + SD PNG packs + on-screen fetch diagnostics are all keepers.

Re your draft disclaimer: I built the **Heltec V4 / LittleFS** env (`heltec_v4_tft_companion_radio_usb_tcp_touch`) on your branch + the latest touch changes — compiles clean, and the SD paths stay correctly `HAS_TDECK_GT911`-gated, so keep that gate (the V4 has no SD slot). It's on `main` now alongside this release's other fixes.

PR #12, opened 2026-06-17
touch UI: battery & telemetry history, map zoom slider + night mode, file image viewer, status-bar tools, Snake

Vybo:
A batch of touch-UI features for the LilyGo T-Deck and, where noted, the Heltec V4 TFT. Each feature is its own commit; the branch builds on both `LilyGo_TDeck_companion_radio_touch` and `heltec_v4_tft_companion_radio_usb_tcp_touch`. SD-backed features fall back to internal SPIFFS or are disabled on the V4 (no SD slot) as noted per feature.

---

## Status bar — microSD activity LED
A small amber dot left of the Wi-Fi glyph, lit briefly on any microSD read/write (tile cache, file-manager browse/copy, log writes). Edge-triggered on the existing UI loop — no new timer. Inert on the V4 (never lit).

### Screenshots
SCR-20260617-nots

## Status bar — async-request spinner
A refresh glyph appears next to the clock while a mesh request is in flight (telemetry, repeater admin, ping, trace, range test, sightline) and clears on reply/timeout. Parked left of the clock so it never overlaps it.

### Screenshots
SCR-20260617-nowt

## Status bar — hold to screenshot
Press-and-hold the status bar for 3 seconds to capture the composited screen to `screenshots/<timestamp>.bmp` on the SD card (16-bit BMP — fastest viewable format; written via a full-screen redraw through the flush hook). The hold suppresses the control-center tap. T-Deck only (needs SD).

## Battery — 24h history chart
Tap the battery in the status bar to open a 24-hour voltage chart. Logged every 5 minutes; prefers the SD card on the T-Deck (`/meshcomod/battery.log`) and falls back to internal SPIFFS on the V4, so the chart works on both builds. Includes an optional CPU-MHz overlay (toggle chip), an estimated-life-left readout from the discharge slope (ignoring charging samples), and a clear-history button. Window scrolls if it overflows the screen.

## Telemetry — per-node history window
Requesting telemetry from a contact opens a per-node window with a chart of battery voltage, temperature and humidity, built from a per-node log (`/meshcomod/telemetry/<id>.log`). A Request button (refresh glyph) sends a poll on demand — the window no longer auto-requests on open. A gear opens a settings panel: show/hide each series (chart redraws live), a manual auto-poll interval (minutes, 0 = off), and clear-history. A time-range selector limits the displayed window. T-Deck only; the V4 keeps the request/reply toast.

## Map — zoom slider
The `+`/`−` overlay buttons are replaced by a single magnifier button that toggles a full-width zoom slider along the bottom of the map (tap again to hide). A "zoom N" readout sits centered above the slider and updates live while dragging. The chosen level is persisted across reboots and takes priority over the first-open auto-snap.

## Map — night mode
A "Night mode (invert)" toggle in Map options inverts tile colours at render time only (the cached tiles are never altered) and re-decodes the visible tiles immediately. While night mode is on, the status-bar text and the bottom tab-bar icons flip to off-white on the map tab so they stay legible over the dark tiles. Persisted.

## Map — zoom + tile-path line
A second line under the © OpenStreetMap attribution shows the current zoom level and the center tile path (`z / x / y`).

## Map — "Show on map" from a contact
The contact action sheet gains a "Show on map" entry that centers the map on that node's last-known GPS position.

## File browser — image viewer
Tapping an image in the file browser opens a read-only viewer with windowed/full-screen toggle. Supports JPEG, PNG (lodepng) and BMP (16/24-bpp) — so screenshots saved by the status-bar hold are viewable on-device. Works on whichever filesystem the browser is showing.

## Chat — hop count on received messages
Received flood messages show their hop count next to the timestamp.

### Screenshots
SCR-20260617-noin

## Contacts — sort unknown last-heard to the end
Contacts whose last-heard time is unknown ("?") sink to the bottom of the list instead of mixing in with recently-heard nodes.

## About — microSD size + fork build note
The storage section shows the microSD card size and free space (T-Deck). The build section tags unofficial fork builds with a note.

## Settings — hide device name
An option to hide the device/profile name in the status bar.

## Apps — Snake game
A self-contained Snake game in the app drawer: canvas playfield sized to the screen, swipe or trackball to steer, a "New game" gate, a pause/resume button, and a close button. The app-drawer grid now scrolls when there are more tiles than fit, instead of shrinking them. Proof of concept for extracting functionality outside of the main UITask class.

Adjusted minimum App tile size for T-Deck as well, so now with the RF Analyzer also present starts scrolling the App page.


platformio.ini
```diff
@@ -146,6 +146,7 @@ build_flags =
   -D DISPLAY_CLASS=ST7789LCDDisplay
   -D ADMIN_PASSWORD='"password"'
   -D MULTI_TRANSPORT_COMPANION=1
+  -D WADAMESH_FORK_BUILD=1
```
Vybo, line 149: Should be changed to 0 before merging.

B0N3-GD:
@Vybo this is huge, awesome work

Vybo:
> @Vybo this is huge, awesome work

@B0N3-GD  Thanks. I have few other things in mind, will submit them when I have time. I'll try to isolate the features to not submit such huge changes at once. Btw, there's a leftover, see my comment above, the fork build flag should be set to 0 for your main repository.

PR #61, opened 2026-06-26
ui-touch: persist the map "Show link lines" toggle across reboots

Yazutsu:
### Problem
In the map options popup, the **Show link lines** toggle did not survive a
reboot (hardware or software) — it always reverted to its default.

### Root cause
The four sibling map-option toggles (Show coordinates, Show tile z/x/y, Show
contacts, Night mode) persist via `TouchPrefsStore`, but the link-lines switch
only set the runtime `s_map_show_links` in `mapOptLinesCb` and never wrote a
pref, so the value was lost on every boot.

### Fix
- Add a trailing `map_show_links` field to `TouchCfg` (cfg **v20**, default `1`
  to preserve the current "links on" behaviour).
- Save it from `mapOptLinesCb` and load it at boot in `UITask::begin()`,
  mirroring the existing map toggles exactly.

Because the field is trailing and the loader is size-tolerant (defaults are
applied first, then the stored blob is copied over), existing configs migrate
cleanly — no settings are reset on update.

### Testing
Built for `LilyGo_TDeck_companion_radio_touch` and flashed on a T-Deck Plus.
Toggled **Show link lines**, rebooted (both soft and hard) — the setting is now
retained. Verified the other map options (coords / tile z-x-y / contacts /
night) still persist correctly across the v19→v20 migration.

### DIFF
--- a/src/helpers/esp32/TouchPrefsStore.cpp
+++ b/src/helpers/esp32/TouchPrefsStore.cpp
@@ -35,7 +35,7 @@ static bool s_begun = false;
 // short read (→ treat as absent → defaults); `ver` lets later builds add fields.
 static const char* KEY_CFG = "cfg";
 static const uint16_t TOUCH_CFG_MAGIC = 0x5743;   // 'WC' (WadaCfg)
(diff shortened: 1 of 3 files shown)


Yazutsu:
Built from current main (commit 83cf020, just past beta_16) and flashed on a T-Deck Plus. Upload was over USB (preserves NVS), so the v19→v20 migration was exercised on an existing config — the previously-saved map options were retained, and Show link lines now persists across reboots too.

B0N3-GD:
Thanks @Yazutsu — this shipped in **beta_18** 🎉

Applied essentially as you wrote it: `map_show_links` added to `TouchCfg` (cfg v20, default 1), saved in `mapOptLinesCb`, loaded in `UITask::begin()` — landed in f6192aa with you credited as co-author, and called out in the [beta_18 release notes](https://github.com/ALLFATHER-BV/wadamesh/releases/tag/beta_18). Really appreciate the clear root-cause write-up and the note about the trailing-field v19→v20 migration. Closing as implemented.

PR #65, opened 2026-06-27
ui-touch: add Polish characters to the Sym picker and accent popups

Yazutsu:
### Motivation
The touch UI had no easy way to enter Polish-specific letters, so Polish text
had to be written without diacritics — which is often ambiguous. This adds the
missing letters through both existing input paths.

### Changes
**Sym special-character picker** (`k_special_items`): add ą ć ę ł ń ś ź ż
(o-acute was already in the list). The count is `sizeof`-derived, so no manual
bookkeeping.

**Accent popups** (`kAccentSets`): append the Polish variant to the relevant
base letters, upper- and lower-case:
- `a → ą`, `e → ę`, `c → ć`, `n → ń`, `z → ż / ź`
- new set for `l → ł`
- new uppercase set `S → Ś / Š`

`o → ó` and `s → ś` were already covered and are unchanged.

### No font change needed
The `extras_*` fallback fonts are already built with Latin Extended-A
(`-r 0x0100-0x017F`), and the `g_font_* → … → extras_*` fallback chain resolves
the glyphs, so every added character renders with no font rebuild. Verified each
codepoint (upper + lower) is present in `extras_font_16.c`.

### Testing
Based on current `main` (`2d49fb6`, just past `beta_18` release),
built for `LilyGo_TDeck_companion_radio_touch` and flashed on a T-Deck Plus.
Verified the accent popups (`l → ł`, `a → ą`, `e → ę`, `c → ć`, `n → ń`,
`z → ż/ź`, plus uppercase), the Sym picker showing the full Polish set, that the
previously-working accents (`o → ó`, `s → ś`) are unaffected, and that inserted
characters render correctly in sent chat messages — not just in the editor.

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -4210,37 +4210,40 @@ static void composerAutoGrowCb(lv_event_t* e) {
 // the extras_* fallback fonts (Latin-1 + Latin-Extended-A), so they render in
 // the textarea and chat. Keys carry LV_BTNMATRIX_CTRL_NO_REPEAT (KeyboardLayouts
 // .cpp) so a hold doesn't auto-repeat — it cleanly long-presses instead.
-static const char* const kAccA[]   = {"à","á","â","ä","ã","å"};
-static const char* const kAccA_u[] = {"À","Á","Â","Ä","Ã","Å"};
-static const char* const kAccE[]   = {"è","é","ê","ë","ě"};
-static const char* const kAccE_u[] = {"È","É","Ê","Ë","Ě"};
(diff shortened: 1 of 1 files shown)


Yazutsu:
It looks like its merged in beta20 allready.

B0N3-GD:
Thanks, @Yazutsu — your Polish characters for the Sym picker and the long-press accent popups are in the shipped firmware. `ą ć ę ł ń ś ź ż` (plus the uppercase `Ś`) are all reachable now, with no font rebuild needed thanks to the Latin Extended-A fallback chain you spotted.

Closing as **shipped**. Much appreciated — and thanks too for the WAV notification sounds (#66). 🙏

PR #66, opened 2026-06-27
ui-touch: custom WAV notification sounds (per-event: message / DM / @-mention)

Yazutsu:
### What
Adds custom notification sounds on the T-Deck, configured in Settings → Sound.
Each event — **message**, **direct (DM)**, **@-mention** — can use the built-in
chime or a user-supplied **WAV** picked from the File Manager.

### Settings → Sound
- Master **Sound** switch, then independent on/off for **Message**, **DM**, and
  **@ mention** (DM previously rode the Messages switch; it now has its own).
- Three sound-file rows. Tapping one offers **Choose .wav from files** (opens the
  File Manager — opening a `.wav` shows **Play** + **Set as notification sound**,
  mirroring the wallpaper picker) or **Built-in (default)**.
- After setting from the File Manager, it returns to Settings → Sound.

### Playback
- New `tdeckPlayWavFile`: streams a PCM **16-bit**, mono/stereo (downmixed) WAV
  from internal flash or the SD card straight to the I2S amp; sample rate is read
  from the header; capped to ~6 s. Unsupported files are rejected at selection,
  and any failure falls back to the built-in chime.
- Built-in chimes are distinct per event (message, DM, @-mention).
- Routing distinguishes DM (`contactMessage`) from channel/room messages.

### Storage
Per-event file paths and the new DM on/off are stored as their own prefs
(`sndf_*` for files — kept separate from the `snd_*` on/off keys).

### Safety
File browsing reuses the File Manager (one directory at a time) rather than a
recursive SD walk — nested directory handles on the shared LoRa SPI bus would
deadlock the UI. V4 (piezo) and Tanmatsu keep the built-in chimes; the WAV
picker is T-Deck only.

### Testing
Based on current `main` (`1e00ac3`, beta_19), built for
`LilyGo_TDeck_companion_radio_touch` and flashed on a T-Deck Plus. Verified:
picking/playing WAVs per event from the SD card, return to Settings → Sound and
clean exit afterwards, the four independent on/off switches, DM vs channel vs
@-mention routing, built-in fallback, and rejection of non-16-bit-PCM files.

### DIFF
--- a/src/helpers/esp32/TouchPrefsStore.cpp
+++ b/src/helpers/esp32/TouchPrefsStore.cpp
@@ -542,10 +542,12 @@ bool touchPrefsSetLockWallpaper(const char* path) {
 }
 
 static const char* soundFileKey(int slot) {
+  // Distinct from the on/off keys snd_msg/snd_dm/snd_men (those are uchar) —
+  // a uchar and a String must not share an NVS key.
   switch (slot) {
-    case TOUCH_SND_DM:  return "snd_dm";
-    case TOUCH_SND_MEN: return "snd_men";
-    default:            return "snd_msg";
+    case TOUCH_SND_DM:  return "sndf_dm";
+    case TOUCH_SND_MEN: return "sndf_men";
+    default:            return "sndf_msg";
   }
 }
 int touchPrefsGetSoundFile(int slot, char* out, int out_cap) {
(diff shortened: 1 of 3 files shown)


B0N3-GD:
Thanks a lot for this, Yazutsu. The custom WAV notification sounds (per-event message / DM / @-mention, with the SD WAV chooser and the DM on/off switch) shipped in beta_20: https://github.com/ALLFATHER-BV/wadamesh/releases/tag/beta_20

You're credited in the release notes and the thanks. Closing as shipped. The Polish input PR (#65) is still open and on my list.

PR #78, opened 2026-06-30
ui-touch: add MP3 player app (Apps -> Music) for T-Deck Plus

Yazutsu:
## MP3 player app (Apps → Music) for T-Deck Plus

Adds a self-contained MP3 player launched from the Apps drawer on the LilyGo T-Deck Plus (ESP32-S3, GT911 touch, MAX98357A I2S amp). It plays `.mp3` files from the SD card.

### Features
- Play / pause / stop, previous / next
- Shuffle, and single / all repeat
- Volume control (shares the system sound volume — see below)
- Settings gear and optional background play
- Elapsed time on the status bar (tap it to reopen the player)
- Remembers the last-used folder and reopens to its first track

The feature itself lives in two new files (`src/ui-touch/Mp3Player.{cpp,h}`). Everything else in the diff is integration into `UITask`; the rationale for each body change is below.

### Why the UITask changes

**Audio backend.** One persistent decode task owns I2S0 for the player's lifetime and is fed new tracks via flags, instead of a task per track — the per-track approach raced the SD and I2S libraries and caused first-track silence plus a use-after-free when switching tracks quickly. A small custom `AudioOutput` (`TdeckMp3Out`) installs I2S0 with the exact config the firmware uses elsewhere and writes non-blocking, so the decode loop keeps polling the stop/pause/volume flags.

**No stutter on the shared SPI bus.** The display, the LoRa radio and the SD card share one SPI bus, so decoding straight off the card stuttered on every incoming packet / screen redraw (the decode task's SD read blocked on the bus). A plain read-ahead buffer does not help: `AudioFileSourceSD` has no real non-blocking read, so the library's `fill()` blocks on the same bus. The fix is `AudioFileSourcePSRAMRing` — a dedicated producer task pours the file into a fixed PSRAM ring (a FreeRTOS StreamBuffer) and the decode task reads only from the ring. When the bus is busy only the producer waits; the decoder keeps draining the ring, so playback never underruns — for files of any size, with no long up-front load. The ring is allocated once for the program lifetime (1 MB, falling back to 512 / 256 KB) to avoid both PSRAM fragmentation and the ~4 MB bank-ceiling allocation failures seen with an earlier per-track buffer.

**Status-bar elapsed time** (`createStatusBar` / `updateGlobalStatusBar`): shown to the left of the clock while playing; tapping it reopens the player. It uses the same `RIGHT_MID` alignment family as the clock so it inherits the clock's exact vertical placement on both the normal and the tall status bar. Aligning relative to the clock instead read a stale, pre-layout clock position on the tall bar and pushed the text off the top.

**Audio priority over notifications** (`tdeckAudioInstallRate` / `tdeckPlayWavFile` / `tdeckNotifyTaskFn`): while the player holds I2S0 (`mp3OwnsI2S()`), the notification chime is suppressed — both the WAV path and the tone fallback — so an arriving message can no longer install over, or zero, the player's live I2S stream mid-playback.

**Overlay dismissal on navigation** (`navGoToMainTab` + HOME hotkey): the player overlay is dismissed on tab navigation and on the T-Deck HOME hotkey, fixing an inherited bug where the overlay could stay on screen after navigating away.

**Stop on lock / screen-off** (`setCpuForScreen`): when the screen turns off (manual lock or idle timeout) the CPU drops to 80 MHz, where MP3 decoding cannot keep up. The player is fully stopped at the 240→80 MHz transition so the CPU can idle cleanly with no garbled audio.

**One shared system volume** (`toggleBuzzer` / `Mp3Player`): the player uses the existing system sound volume (`snd_vol`) for playback and persists changes back to it, so the player volume, the Settings → Sound +/- buttons and the Control Center slider are a single saved setting shared with notifications, rather than a second, non-persisted player-only volume.

**Master sound switch stops music** (`toggleBuzzer` + `uiSoundMuted`): turning the master "Sound" switch off — via Settings → Sound, the Control Center toggle, or the long-press VOL+ gesture, all of which route through `toggleBuzzer()` — now also stops playback, and `play()` is gated while muted, so the player respects the master switch.

**App registration:** adds the Music tile to the Apps drawer and an `.mp3` branch to the File Manager so a tapped track opens in the player. `platformio.ini` gains the ESP8266Audio dependency used for MP3 decoding.

### Testing
Built and flashed on the T-Deck Plus (`LilyGo_TDeck_companion_radio_touch`, on top of beta_21). Verified: playback, transport and volume; no stutter with incoming DMs for both small and large files; correct status-bar time on normal and tall bars; overlay dismissal on navigation; full stop on manual and idle lock; single shared volume persisted across reboots; master-sound-off stops music; last folder remembered and reopened to its first track.

Single DCO-signed commit.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -175,6 +175,7 @@ build_src_filter =
   +<../variants/heltec_v4/*.cpp>
 
 lib_deps =
+  earlephilhower/ESP8266Audio @ ^1.9.7
   ; MeshCore core = the meshcomod monorepo built as a lib, pinned by tag.
   ; Touch-app files (TouchPrefsStore, WifiRuntimeStore, transports, ...) are
   ; vendored in this app's src/ and dropped from the lib via -DMC_VENDORED_TOUCH_APP.
@@ -349,6 +350,7 @@ build_src_filter =
   +<../variants/lilygo_tdeck/*.cpp>
 
 lib_deps =
+  earlephilhower/ESP8266Audio @ ^1.9.7
   https://github.com/ALLFATHER-BV/meshcomod.git#core-v1.16.4
   SPI
   Wire
--- a/src/ui-touch/Mp3Player.cpp
+++ b/src/ui-touch/Mp3Player.cpp
@@ -1,0 +1,736 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+#include "Mp3Player.h"
+#include <driver/i2s.h>          // I2S amp — install with the firmware-matched config
+#include "../helpers/esp32/TouchPrefsStore.h"   // remember last track (blob prefs)
+
+#if defined(HAS_TDECK_GT911)   // player needs the I2S amp + File Manager (T-Deck)
+
+#include <Arduino.h>
+#include <SD.h>
+#include <SPIFFS.h>
+#include <string.h>
+
+#include "AudioGeneratorMP3.h"
+#include "AudioFileSourceSD.h"
+#include "AudioFileSourceSPIFFS.h"
+#include "AudioOutput.h"
+#include <esp_heap_caps.h>
+#include "freertos/FreeRTOS.h"
+#include "freertos/stream_buffer.h"
+
+// ---------------------------------------------------------------------------
+// Streaming source with a BACKGROUND FILL TASK.
+// Every SD read happens on a dedicated producer task that pours bytes into a
+// large PSRAM ring (FreeRTOS StreamBuffer). The decode task only ever pulls
+// from the ring, so an SPI-busy window (incoming LoRa packet, display flush)
+// stalls just the producer -- the decoder keeps draining the ring and never
+// underruns. Works for files of ANY size with a fixed ring, and removes the
+// long whole-file load (playback starts as soon as the first chunk lands).
+// ---------------------------------------------------------------------------
+class AudioFileSourcePSRAMRing : public AudioFileSource {
+ public:
+  AudioFileSourcePSRAMRing(AudioFileSource* under, uint8_t* storage, uint32_t storageBytes)
+      : under_(under) {
+    size_ = under_ ? under_->getSize() : 0;
+    if (storage && storageBytes > 64)
+      ring_ = xStreamBufferCreateStatic(storageBytes - 1, 1, storage, &ringCtl_);
+    if (ring_ && under_)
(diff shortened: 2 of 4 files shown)


B0N3-GD:
Hey @Yazutsu  going to check how to bring this in once I've reached some core functionality stability

richrd:
I'd love an MP3 player on my T-Deck, but personally I'd rather install it as a separate firmware. I'm afraid adding too many features to Wadamesh which are unrelated to the Mesh will bloat the firmware and it'll grow in size, making it harder to fit multiple firmwares on one device 🤔

B0N3-GD:
> I'd love an MP3 player on my T-Deck, but personally I'd rather install it as a separate firmware. I'm afraid adding too many features to Wadamesh which are unrelated to the Mesh will bloat the firmware and it'll grow in size, making it harder to fit multiple firmwares on one device 🤔 

I agree @richrd mesh focus and then we'll see. Especially now since there are some importsnt things to get stable first.

Yazutsu:
Some numbers in case they help the decision — measured on beta_30, `LilyGo_TDeck_companion_radio_touch`, stock `main` vs this branch rebased on top of it:

| | Flash | Static RAM |
|---|---|---|
| beta_30 stock | 2,975,733 B (73.2%) | 102,440 B (31.3%) |
| beta_30 + player | 3,099,401 B (76.3%) | 102,680 B (31.3%) |
| **delta** | **+120.8 KB (+3.1%)** | **+240 B (~0)** |

That leaves **~941 KB (23.7%) of the app partition free** with the player in. The flash cost is mostly the Helix MP3 decoder pulled in via ESP8266Audio.

Runtime memory is lazy by construction:
- **Never opened → nothing allocated.** Just the code in flash and a few bytes of statics.
- **While playing:** ~40 KB decode-task stack + ~25–30 KB decoder state on the internal heap, freed on stop.
- **PSRAM:** a 1 MB streaming ring allocated once at first playback and kept for the program lifetime (deliberate — avoids re-allocation fragmentation), out of the ~8 MB the firmware barely touches otherwise.
- Everything is gated behind `HAS_TDECK_GT911`, so the Heltec/Tanmatsu builds pay nothing.

To be clear: **no pressure on merging this** — I get that a music player is a judgement call for a mesh firmware. I'm running it daily myself either way, so I'll simply keep the branch rebased onto `main` so it stays a clean, conflict-free merge whenever (and if) you decide it fits.

B0N3-GD:
going to close for now based on the previous discussions.

PR #85, opened 2026-07-03
ui-touch: keyboard-backlight brightness control for the T-Deck (issue #84)

### LINKED ISSUE #84: feature request: Keyboard lighting adjustment
So it would be a nice option to be able to set keyboard a little dimmer than original that I believe is 100%

Yazutsu:
## Keyboard-backlight brightness control for the T-Deck

Fixes #84.

The T-Keyboard's C3 firmware accepts a 2-byte I2C command `[0x01, level 0-255]`, but the level the UI drove it at was hardcoded `0xFF` — so the backlight was always 100% when lit. This adds a **Keyboard backlight** control at the top of Settings → Keyboard: a live slider plus Off / 25 / 50 / 75 / Max one-tap presets. The chosen level applies to the *on* and *auto* modes and to the new-message flash; `0` keeps the keyboard dark even when the mode is on.

The percent persists in the existing `kbd_bl` pref (shared with the Tanmatsu Keys slider — no new config fields, no TouchCfg version bump), and adjusting the control while the mode is *off* flips it to *on*, mirroring the Tanmatsu control-centre behaviour.

### Latency notes (why the diff touches the input helpers)

Three things made the first cut of this control feel bad, and each fix is in this PR:

- **Perceptual mapping.** Perceived LED brightness is roughly logarithmic in PWM duty, so a linear pct→duty map felt coarse at the dim end (small steps = big visible jumps). The percent is now gamma-2 mapped (`duty ∝ pct²`), which spends most of the control range on the dim end; any nonzero percent still guarantees a visibly lit level (≥1).
- **NVS writes off the hot path.** The pref setter rewrites the whole TouchCfg blob via `cfgFlush()`, which can block the UI thread for a good fraction of a second — far too slow per tap. The save is now debounced: a one-shot timer re-armed on each change writes once after the user stops adjusting.
- **I2C flush cadence.** The backlight write must stay on the touch-poll task (the keyboard shares the I2C bus with the GT911), but it used to ride the ~32 ms keyboard-scan divider — and the Settings control additionally waited for the next `UITask::loop()` pass to notice the change. Requests are now made directly by the control and flushed by a new `tdeckKeyboardFlushBacklight()` on every poll tick (~8 ms; a single flag check when idle), so the control feels as immediate as the screen-brightness slider, whose `ledcWrite` hits the ESP32's own PWM peripheral synchronously.

### Testing

Built and flashed on a T-Deck Plus (`LilyGo_TDeck_companion_radio_touch`, on top of beta_26). Verified:  live dimming while dragging (the C3 does PWM the intermediate levels); presets jump instantly; the value survives a reboot; the on/off/auto modes and the message flash honour the chosen level; typing latency unaffected (key scan cadence unchanged).

Single DCO-signed commit.

### DIFF
--- a/src/helpers/input/TDeckKeyboard.cpp
+++ b/src/helpers/input/TDeckKeyboard.cpp
@@ -34,9 +34,9 @@ void tdeckKeyboardSetBacklight(uint8_t level) {
   if (level != s_bl_desired || !forced) { s_bl_desired = level; s_bl_dirty = true; forced = true; }
 }
 
-void tdeckKeyboardPoll() {
-  if (!s_inited) return;
-  if (s_bl_dirty) {
+void tdeckKeyboardFlushBacklight() {
+  if (!s_inited || !s_bl_dirty) return;
+  {
     s_bl_dirty = false;
     // LilyGo T-Keyboard backlight: 2-byte command [0x01, brightness] (0 = off).
     Wire.beginTransmission(PIN_KB_ADDR);
@@ -44,6 +44,11 @@ void tdeckKeyboardPoll() {
     Wire.write(s_bl_desired);    // 0 = off, 1-255 = brightness
     Wire.endTransmission();
   }
+}
+
+void tdeckKeyboardPoll() {
+  if (!s_inited) return;
+  tdeckKeyboardFlushBacklight();
   if (Wire.requestFrom((int)PIN_KB_ADDR, 1) != 1) return;
   uint8_t key = Wire.read();
   if (key == 0) return;                       // no key this scan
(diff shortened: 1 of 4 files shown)


B0N3-GD:
Merged, thanks for the thorough writeup. Reviewed the diff and built it clean on all three targets (LilyGo T-Deck, Heltec V4, Tanmatsu P4), so the HAS_TDECK_KEYBOARD gating holds on the non-keyboard boards. Nice touches on the gamma-2 curve, the debounced NVS write (keeps cfgFlush off the UI thread), and moving the flush onto the touch-poll tick without changing the key-scan cadence. This will ship in the next beta.

PR #88, opened 2026-07-03
ui-touch: make the Settings keyboard-backlight Off preset switch the mode off

Yazutsu:
quick logic fix

Follow-up to the keyboard-backlight brightness control (issue #84). The Off preset used to set the brightness percent to 0 while leaving the backlight MODE untouched, inventing a second dark state: the keyboard went dark but the control-center  - which cycles and shows the mode - kept saying on/auto.

Off now switches the mode off, the same axis the CC  shows, so the two surfaces can never disagree. The brightness slider becomes 1-100: dark is only expressible via the mode, and the on/auto modes always glow visibly at the chosen level. The stored percent is left untouched by Off, so re-enabling the backlight restores the previous brightness; a stored 0 from the previous build is clamped to 1 at boot.


B0N3-GD:
Merged, thank you. Good catch on the two dark states, and the mode is exactly what the control center chip shows, so they stay in sync now.

B0N3-GD:

PR #123, opened 2026-07-07
fix(esp32): honor use_sd from touch prefs at boot

scratchdiver:
## What

Fixes the **Store data on SD** toggle being ignored after reboot on ESP32 touch builds.

The Settings UI persists `use_sd` via `SdNvsPrefs` (written to `/prefs/touch.kv` when NVS is unavailable or when using SD-backed prefs), but `main.cpp` only read the raw NVS `Preferences` key at boot. After enabling SD storage in the UI, a reboot could still show `[BOOT] storage: SPIFFS` and keep contacts on SPIFFS.

## Why

On a large mesh, full `contacts3` rewrites on SPIFFS can block for several seconds and contribute to UI freezes. SD storage is much faster for these writes, but only when boot actually selects SD. This bug meant users could enable **Store data on SD** and still boot into SPIFFS after reboot — so the performance benefit of SD never applied even when the toggle appeared set.

## Fix

- `touchPrefsSetUseSdStorage()` now dual-writes: existing `SdNvsPrefs` path **and** raw NVS `use_sd`.
- New `touchPrefsReadUseSdAtBoot()` reads NVS first, falls back to parsing `/prefs/touch.kv`, and syncs NVS when the file has the authoritative value.
- Boot path in `main.cpp` uses `touchPrefsReadUseSdAtBoot()` instead of `Preferences.getBool("use_sd")`.

## Scope / not touched

- Chat UI, contacts save performance, storage write tracing — separate work.
- No TouchCfg version bump; `use_sd` remains a standalone pref key as before.

## Testing

- [x] Built `LilyGo_TDeck_companion_radio_touch`
- [x] Built `heltec_v4_tft_companion_radio_usb_tcp_touch`
- [x] T-Deck: enable **Store data on SD** in Settings → reboot → boot log shows SD storage selected
- [x] Toggle off/on + reboot: preference survives

Single DCO-signed commit.

### DIFF
--- a/src/helpers/esp32/TouchPrefsStore.cpp
+++ b/src/helpers/esp32/TouchPrefsStore.cpp
@@ -6,6 +6,8 @@
 
 #include "SdNvsPrefs.h"   // NVS, or SD /meshcomod fallback when NVS is unusable (Launcher)
 
+#include <Preferences.h>
+#include <SPIFFS.h>
 #include <stddef.h>   // offsetof
 #include <string.h>   // memcpy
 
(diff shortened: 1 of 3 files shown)


B0N3-GD:
Merged, thank you. Good find and a clean fix, this was exactly why the SD toggle kept reverting on Launcher installs. Two small edge cases we noticed in review (the setter return value on Launcher, and turning SD off again when NVS is unusable) are being handled in a follow-up commit so you do not need to do anything.

PR #125, opened 2026-07-07
ui-touch: GPS position as the first quick-reply row

Yazutsu:
Sharing where you are is one of the most common things typed into a mesh chat, and until now the only way was to hand-copy coordinates off the GPS screen. This adds the node's live position as the **first row of the chat quick-reply picker**:

- **With a live fix** the row shows the coordinates right on the button (`⌖ 52.41234, 21.03456` — `%.5f`, the same format the firmware's GPS screens use). Tapping it inserts the plain coordinates into the composer **at the cursor, with no auto-send** — the same contract as the macro rows (including the clear-selection guard), so the operator can add context before sending.
- **Without a fix** the row is greyed out and `LV_STATE_DISABLED`, so it can't fire. The state is evaluated each time the picker opens.

The picker card grows by exactly one row height (the card-height formula gains a `+1`), keeping the existing fits-under-the-status-bar sizing intact on both the small boards and the Tanmatsu panel.

Inserted text is deliberately the bare `lat, lon` — no prefix, no maps URL — to keep the payload minimal for LoRa and let the sender phrase the message.

### Testing
Built and flashed on a T-Deck Plus (`LilyGo_TDeck_companion_radio_touch`, on top of beta_36). Verified: coordinates shown live on the row and inserted at the cursor without wiping typed text; row greyed and inert without a fix (including right after boot, before first fix); state refreshes on picker reopen; the 7-row card fits under the status bar; no interaction with the new beta_36 field edit menus.

Single DCO-signed commit.

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -6039,6 +6039,23 @@ static void qrPickCb(lv_event_t* e) {
 #endif
 }
 
+// First picker row: insert the node's current GPS position (issue-free text, no
+// auto-send - same contract as the macro rows). Only clickable with a live fix.
+static void qrGpsPickCb(lv_event_t* e) {
+  if (lv_event_get_code(e) != LV_EVENT_CLICKED) return;
+  LvChatPanel* p = s_qr_panel;
+  closeQuickReplySheet();
+  if (!p || !p->composer_ta) return;
+#if defined(ESP32)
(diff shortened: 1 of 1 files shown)


B0N3-GD:
Merged, thank you. Clean insert contract and the no-fix greyout works as described. This will ship in the next test build.

B0N3-GD:

PR #130, opened 2026-07-07
touch: virtualize chat scrolling (bubble + compact)

### LINKED ISSUE #126: Virtualize channel message display to allow more than 30 message history
The channel message display has a hard-coded limit of 30 messages (60 in compact mode). For popular channels like Public the 30-message window can limit the viewable history to the last hour or so — nothing earlier will be displayed even if the internal store has history of 500 messages going back over a day or more.

The channel messages display should instead be virtualized: only materialize LVGL bubble widgets for the visible viewport plus a small overscan band (~120 px above and below). The count is not fixed — it varies with message height — but on a T-Deck that typically works out to on the order of **10–15 bubbles** in memory at once, rather than 30 full widget trees. As the user scrolls further back, newer bubbles are torn down and older ones are paged in, allowing scrolling back through the full stored history.

This smaller on-screen window should also reduce RAM pressure indirectly: each chat bubble is an LVGL object tree (labels, padding, styles, event handlers), so keeping fewer materialised at once leaves more headroom for the rest of the UI, mesh stack, and Wi-Fi/BLE buffers — especially on the 320 KB internal DRAM budget.

I actually have this coded already and working apparently well on T-Deck Plus, along with a couple of small UI changes to support it (e.g. a matching button to jump to the oldest message), but I wanted to gauge interest from the contributors before I do the work of merging the change in a PR.

If you think this change is aligned with the goals of the project please leave a comment here in support.

scratchdiver:
## Summary

- Virtualize open-chat message rendering: only the visible viewport plus ~120 px overscan is materialised as LVGL widgets, while layout offsets cover the full stored thread (up to `msgCap()`). On T-Deck this is typically ~10–15 bubbles in memory instead of the previous 30/60 cap.
- Supports both bubble mode and **Compact messages (IRC style)** from Settings → Display, including day separators ("Today", "Yesterday", …) and even compact-row spacing.
- Add jump-to-oldest button (paired with existing jump-to-latest) and compressed LVGL scroll coordinates for very long histories.

Fixes #126

## Why

Popular channels can store hundreds or thousands of messages on SD, but the UI previously capped rendered history at 30 bubbles (60 compact). Virtualization keeps scroll performance and RAM headroom while letting users scroll back through the full stored history.

## Test plan

- [x] T-Deck Plus (`LilyGo_TDeck_companion_radio_touch`) builds cleanly
- [x] Open a busy channel — scroll back through history beyond the old 30-message window
- [x] Jump-to-oldest / jump-to-latest buttons work and hide appropriately
- [x] Toggle **Compact messages (IRC style)** — rows virtualize with even spacing and alternating stripe
- [x] Day separators appear when scrolling across calendar days
- [x] Unread "New" divider still appears on open when there are unread messages


scratchdiver:
investigating stability of crash fixes

scratchdiver:
ok I have a stable commit for this now but the work exposed a couple issues related to message store persistence that should get fixed first, issue here: https://github.com/ALLFATHER-BV/wadamesh/issues/134
and fix here: https://github.com/ALLFATHER-BV/wadamesh/pull/135
I will make this PR ready for review when the above issue is merged first

scratchdiver:
Rebased onto `main` after #135 landed and squashed to a single commit.

- Includes soak-tested LVGL crash fixes (async refresh/reflow, sync delete in render path, `msgCap()`-sized index buffer for the SD deep ring, open-to-bottom/jump-button fixes).
- Persistence/flush changes stay in #135 — not duplicated here.

scratchdiver:
ready for review & merge

B0N3-GD:
Thanks for the rebase and squash, and for the soak testing notes. This is next in the review queue. Virtualization touches the hottest path in the UI, so it gets a proper deep read rather than a quick skim; I will come back with a full review rather than piecemeal comments.

scratchdiver:
thanks for the update. While working on this I ran into quite a few LVGL crashes due to memory accessed after released, memory freed with wrong api, memory accessed in tap handlers that was already deleted etc. (some of these no doubt came from the virtualization) so there's a lot of things moved to deferred rendering now. I think it is still responsive. I look forward to your feedback.

B0N3-GD:
Full review done: the complete diff, two deep passes (one on object lifetime and memory safety, one on functional correctness), plus builds of all four targets with the PR merged onto current main. Short version: this is genuinely impressive work and we want it in. The deferral architecture (timer, then async, then sync deletes outside indev and refresh context) matches every hard-learned LVGL rule in this codebase, the coordinate compression under the 8191 px limit is clever and self-consistent, and open-at-latest plus the mention-jump actually got better than the old code (the old path could only jump within the last rendered rows). It is not mergeable quite yet though. Here is everything, blockers first.

## Merge blockers

**1. It puts the Tanmatsu over its app-load ceiling.** The P4 launcher's ELF loader caps the app image at 50 MMU pages (3,276,800 bytes). Current main builds at 3,270,192; with this PR it is 3,278,208, which is 1,408 bytes over, and the result is not a build error but an app that silently fails to launch (blue splash, black screen) for every Tanmatsu user. Your trace scaffolding is properly compiled out, so this is ~8 KB of real code against a budget that had ~6.6 KB left. Options: shave the footprint (the compact and bubble paths share a lot of near-duplicate measure/create logic that might fold), or we free embedded data on our side to make room. Happy to coordinate on this one since the ceiling is our platform constraint, not something you could have known.

**2. Confirmed use-after-free: the divider async-delete in `chatVirtReset` races the synchronous cleans.** `chatVirtReset` queues `lv_obj_del_async(s_chat_virt.divider)` (a raw pointer, LVGL 8.3 does no validity check when it fires). Both empty-thread paths in the new `refreshChatDetail` then call `lv_obj_clean(p.msgs)` two lines later in the same callback, which frees the divider synchronously since it is a child of the scroller. The queued async then runs `lv_obj_del` on freed memory at the next timer drain. Repro shape: have the unread divider on screen, then "Reset chat history" on that thread. There is also a cross-panel variant: `openThreadDetailByIdx` does not close the other panel, `refreshChatDetailAsyncCb` refreshes both in one callback, and `chatVirtReset` ignores its argument and acts on the global divider pointer, so panel A's reset can queue the async delete of panel B's divider right before B's sync purge frees it. Suggested fix: drop the `lv_obj_del_async` from `chatVirtReset` entirely and just null the pointer. The divider is always a child of `p->msgs` and every downstream path (clean, sync purge, window clear) already deletes children and nulls the static, which also kills the cross-panel variant.

**3. Compact mode: the per-message long-press menu is dead.** `chatVirtCreateCompactRow` registers `bubbleLongPressMenuCb` for `LV_EVENT_PRESSED`, `RELEASED`, `PRESS_LOST` and `DELETE`, but the callback's first line filters everything except `LV_EVENT_LONG_PRESSED`, which is never registered. So in compact mode, long-press does nothing: Copy, Info, Resend, Ack, Mention and Block are all unreachable. The bubble path still registers `LONG_PRESSED` correctly. It looks like a manual press-tracking rewrite of the callback was intended and did not make the diff; if so, the simple fix is registering `LV_EVENT_LONG_PRESSED` like the bubble path does (the `DELETE` registration does no cleanup and can go too).

**4. Bubble mode: tap-to-resend on a failed message was dropped.** The old bubble path attached `bubbleRetryTapCb` on `LV_EVENT_CLICKED` for `DELIV_FAILED` messages and the footer read "tap to resend"; the new `chatVirtCreateBubble` attaches only the long-press menu and the footer is now just the cross. The compact path kept the retry tap. Net effect with 3: bubble users lose one-tap resend, compact users lose the whole menu. Both regress shipped behavior.

**5. Same-count content changes never rebuild the layout, and the `dirty_timeline` rescue branch is dead code.** The rebuild gate is `panel != p || n != cap || !offsets || compact changed`, with an `else if (g_lv.dirty_timeline)` rescue. But `UITask::loop` clears `dirty_timeline` synchronously right after scheduling the async refresh, and the refresh callback runs on a later timer pass, so it always reads false. Two real consequences. First, once the message ring reaches capacity (500, or 5000 on SD), every new message evicts an old one and n stays constant, so `msg_idx[]` is refreshed but `offsets[]`, `day_sep_y[]` and the total height still describe the previous set: every row renders the next message's text at the previous message's offset, compounding per arriving message until the chat is reopened. A full ring is the steady state of a long-running device, so this is not an edge case. Second, switching threads while the same panel is open (mention-jump, open-DM-from-contact) with equal message counts renders thread B at thread A's offsets. Fix direction: give `need_layout` a thread-identity term (capture the active thread in `s_chat_virt`) and a content-generation term (a ring write counter, or compare first/last `msg_idx`), instead of the dead flag read.

**6. Full-thread re-measure per arriving message.** `chatVirtRebuildLayout` measures the entire thread (two or three `lv_txt_get_size` per message plus a struct copy) on every open and on every received message while the chat is open, since n changes trigger a rebuild. At 500 messages that is tens of milliseconds on the UI thread; on the 5000-message SD ring it plausibly exceeds 100 ms per incoming message. An incremental append for the new-tail case (extend `offsets[]` only) would restore the old cost profile and, done right, also fixes the ring-at-capacity half of 5.

## Should-fix, not blocking

- `chatVirtRenderTimerCb` sets `s_chat_virt_render_async_busy = true` and ignores `lv_async_call`'s return. If that allocation ever fails, the flag never clears and chat rendering wedges until reboot. One line: clear the flag and panel if the call returns anything but `LV_RES_OK`.
- On the keypad-nav boards, `navMoveDir`'s fallback `lv_group_focus_obj(s_nav_objs[0])` has no validity guard, and this PR makes bubbles recycle on every scroll-end, so a stale first slot is much more likely than before. Either guard it with `lv_obj_is_valid` or mark `p->msgs` with the nav-skip flag so bubbles stay out of the focus group (which also stops focus churn during reflows).
- Opening or switching chats now shows one frame of the previous chat's bubbles, then a blank frame, then content (three deferred passes). A synchronous purge of the old children on close or at open-start would kill the flash. Related: a long fast drag past the overscan shows blank rows until scroll-end materializes them; if that is the intended trade, fine, but worth a comment.
- Compact rows update delivery/echo glyphs inline in the wrapping label; if the extra characters push a wrap, the row grows against a stale measured height and overlaps its neighbor until the next rebuild. Bubble mode is immune (footer is a dedicated line). Falls out of the fix for 5 if heights are re-measured on content change.
- `chatVirtClearAllChildren` and the async path of `chatVirtDeleteMsgsChild` are dead code today, and would become blocker-2-class hazards if ever called alongside the sync purges. Suggest deleting both.
- The `MyMesh.h` change (default `FIRMWARE_BUILD_DATE` string) looks like a straggler from a local build and can drop out of the PR.

## What checked out clean, so you do not re-audit it

All five async sites were traced: the render async and refresh async capture no pointers (statics plus panel identity re-checked at fire time), the one-shot timer's self-delete is safe on LVGL 8.3.11, and all seven former synchronous `refreshChatDetail` call sites are converted with no stragglers. The sync-delete discipline holds everywhere (indev wait-release before mass deletes, purges only in async or timer context, the scroller itself never deleted). Window math is right at both edges including empty, single-message and shorter-than-viewport threads. Divider and day separators produce no duplicates and position correctly in both coordinate modes. New messages while scrolled up do not yank the view below ring capacity. Both render modes are fully virtualized and the compact toggle rebuilds properly. Every coordinate handed to LVGL stays under `LV_COORD_MAX`, consistent with keeping `LV_USE_LARGE_COORD` off, and your lv_conf note documenting why is appreciated.

Once the blockers are addressed we will run an on-device soak on a T-Deck with a deep SD history (long scrolls, live updates while scrolled up, compact toggle, resend paths, reset-history with the divider visible) before merging. The crash fixes you folded in are exactly the class of bug we have been chasing on main, so there is real appetite here; thanks for the patience with a review this heavy.

B0N3-GD:
Rather than another round trip, I pushed the review fixes onto your branch (two commits, your history untouched):

f04e017^ fixes the blockers from the review: the divider UAF in chatVirtReset (the reset only nulls the pointer now, the downstream cleans own the widget, which also kills the cross-panel variant), the compact-row menu (it now registers LV_EVENT_LONG_PRESSED, the only event the callback acts on), bubble tap-to-resend restored with the footer hint, a content-generation guard on need_layout (endpoint ring slots, so ring rotation at capacity and equal-count thread switches force a proper relayout) with the dead dirty_timeline rescue removed, a chatVirtTryAppendLayout fast path so a new message extends the offsets tail instead of re-measuring the whole thread, the lv_async_call busy-flag guard, the nav-focus validity guard, the dead ClearAllChildren/DeleteMsgsChild pair removed, and the MyMesh.h build-date straggler reverted.

f04e017 resolves the size blocker at the platform level: the Tanmatsu app was building at -Og this whole time. At -Os it is 2,980,688 bytes, 296 KB under the load ceiling with your full PR in, so nothing of yours needs shaving after all. That one needs a P4 boot check on our side before merge since the board has never run an -Os build.

All three build targets are green with the fixes, and the branch is flashed on a T-Deck for the on-device soak now. Please look over my changes to your code, especially the append fast path and the ring-slot guard, and push on top if you see something off. Once the soak and the P4 boot check pass this merges.

B0N3-GD:
Merged and shipped: this is the headline of beta_38 on the test channel, credited in the release notes and the release body. https://github.com/ALLFATHER-BV/wadamesh/releases/tag/beta_38

The two commits I pushed onto your branch (the review fixes and the -Os build switch for the P4) are described a few comments up; if anything there reads wrong to you, follow-ups are very welcome. The maintainer soaked the branch on a T-Deck through a full day of chat use before the merge, and the P4 boot-verified at -Os with your virtualization in. Thanks for a serious piece of work and for the patience with the heavy review.

scratchdiver:
thanks for the blocker fixes!

PR #135, opened 2026-07-08
touch: fix SD message history flush reliability

### LINKED ISSUE #134: SD chat history (ui_msgs_v1.bin) not flushing; crash can corrupt message store
## Summary

On T-Deck (and other touch builds using the SD deep message ring), incoming chat messages update thread metadata on disk but the message body file (`/meshcomod/ui_msgs_v1.bin`) often does not. After a panic reboot, messages that were visible in RAM (and reflected in unread counts on the home screen) are missing. A crash during a flush can also leave a truncated or corrupt `ui_msgs_v1.bin`, causing the loader to quarantine the file on next boot and present empty channels.

## Expected behavior

- New messages are written to `ui_msgs_v1.bin` on the normal debounced flush cadence (~30s on SD).
- A panic reboot loses at most the messages since the last successful flush.
- A crash mid-write does not destroy the previous good snapshot.

## Actual behavior

During soak testing on a T-Deck with SD storage:

1. **Flushes never complete** unless something else has spawned the core-0 `tile_fetch` worker (map tiles, Wi-Fi scan, LOS, etc.). `flushHistoryIfDue()` arms `s_hist_flush_req` and clears `_msgs_dirty`, but nothing writes the snapshot if the worker task does not exist.
2. **Thread metadata and message bodies diverge.** `ui_threads_v1.bin` updates on a ~1.5s cadence (unread counts, `last_ts`), while `ui_msgs_v1.bin` can remain hours stale. The UI shows new unreads; after reboot only old messages load.
3. **Crash during write corrupts the store.** `uiWriteMsgsFile()` opens `ui_msgs_v1.bin` with `"w"` (truncating immediately). A panic mid-write produces a bad header or truncated file. On boot, `loadMsgsFromStorage()` quarantines the corrupt file → all channels appear empty.
4. **False success on empty snapshot.** `uiHistWorkerFlush()` returned `true` when no snapshot was armed, so a failed arm could look like a successful write.

## Evidence

Soak test (2026-07-07): `ui_threads_v1.bin` timestamp ~4pm; `ui_msgs_v1.bin` stuck at 10:08 AM despite afternoon traffic. After crash, crash-folder copy showed a valid 1.1 MB file (5000-slot SD ring format, `count=512`) — persistence *can* work when the worker runs, confirming the write path rather than the on-disk format is the problem.

## Proposed fix

1. Call `ensureTileFetchTaskRunning()` when arming or waiting on a message-history flush, so the existing worker always picks up `s_hist_flush_req`.
2. Fall back to synchronous `saveMsgsToStorage()` on the loop thread if the worker cannot spawn (rare), instead of clearing `_msgs_dirty` and dropping data.
3. Write to `ui_msgs_v1.bin.tmp` and atomically rename on success (same pattern as `battery.log`).
4. Return `false` from `uiHistWorkerFlush()` when no snapshot is armed, so the retry path runs.

## Scope / non-goals

This issue is about **flush reliability and crash-safe writes**, not the full-ring rewrite strategy. The current design rewrites the entire 5000-slot (~1.1 MB) file per flush; optimizing that (dirty-slot flush, etc.) is a separate follow-up.

## Environment

- Board: LilyGo T-Deck Plus (`LilyGo_TDeck_companion_radio_touch`)
- Storage: SD card, `/meshcomod/ui_msgs_v1.bin`
- Ring capacity: `MAX_UI_MESSAGES_SD` (5000 slots)

scratchdiver:
## Summary

- Ensure the `tile_fetch` worker is running before arming or waiting on a message-ring flush (`s_hist_flush_req` was often never serviced).
- Fall back to synchronous `saveMsgsToStorage()` if the worker cannot spawn, instead of clearing `_msgs_dirty` and dropping data.
- Write `ui_msgs_v1.bin` via `ui_msgs_v1.bin.tmp` and atomic rename (same pattern as `battery.log`) so a panic mid-write does not truncate the live file.
- Return `false` from `uiHistWorkerFlush()` when no snapshot is armed so the retry path runs.

Fixes #134

## Test plan

- [x] T-Deck soak: messages persist across reboot after 60s+ uptime (ui_msgs_v1.bin mtime advances)
- [x] Thread unread counts and message bodies stay in sync after flush
- [x] Panic during heavy chat use no longer quarantines ui_msgs_v1.bin on next boot (atomic replace)
- [ ] Verify on Heltec V4 TFT (SPIFFS path unchanged except worker kick + flush retry semantics)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -157,6 +157,7 @@ constexpr uint16_t k_ui_history_min_version = 6;   // v4/v5 used 96-char records
 // reducing flash write pressure when the ring is large.
 constexpr const char* k_ui_threads_path = "/ui_threads_v1.bin";
 constexpr const char* k_ui_msgs_path    = "/ui_msgs_v1.bin";
+constexpr const char* k_ui_msgs_tmp_path = "/ui_msgs_v1.bin.tmp";
 constexpr uint32_t k_ui_threads_magic   = 0x55495448;  // "UITH"
 constexpr uint32_t k_ui_msgs_magic      = 0x55494D53;  // "UIMS"
 
@@ -34326,6 +34327,10 @@ void UITask::flushHistoryIfDue(unsigned long now) {
   }
   if (_msgs_dirty && now >= _next_msgs_flush_ms) {
     if (s_hist_flush_busy || s_hist_flush_req) {
+      // The worker only exists while the tile-fetch task is spawned (map / Wi-Fi
+      // scan / LOS / …).  Without this kick a pending flush can sit armed forever
+      // while ui_threads_v1.bin keeps updating — messages live only in RAM.
+      ensureTileFetchTaskRunning();
       _next_msgs_flush_ms = now + 1000;   // one flush in flight; new messages ride the next one
       return;
     }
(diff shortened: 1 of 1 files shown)


B0N3-GD:
Merged, thank you. The worker kick was a real hole (the task idles out after 5 seconds, so an armed flush could sit forever), and the tmp plus rename write closes the truncation-on-panic case nicely. Verified building on all three boards.

B0N3-GD:

PR #136, opened 2026-07-08
feat: RAK WisMesh Tap V2 (RAK3312) board support

Ethan-chen1234-zy:
## Summary

Adds board-level support for the **RAK WisMesh Tap V2** (RAK3312 module, ESP32-S3).

New PlatformIO env: `rak_tap_v2_companion_radio_touch`

This is a **hardware glue port only** — display driver, touch input, dual-SPI radio bus, USB flash behaviour, and BOOT-button power menu. No UI screens or `UITask` logic were added or redesigned; only the existing `HAS_RAK_TAP_V2` branches needed by other boards (display class, rotation override, draw-buffer width, power-menu download entry).

## Hardware

| Item | Detail |
|------|--------|
| Module | RAK3312 — ESP32-S3, 16 MB Flash (DIO), 8 MB Octal PSRAM |
| Display | ST7789 240×320 native; landscape 320×240 via LovyanGFX HW rotation |
| Touch | FT5x06 capacitive, I2C 0x38, SDA=9 / SCL=40 / INT=39 |
| LoRa | SX1262 on **HSPI** (SCLK=5, MISO=3, MOSI=6, CS=7, DIO1=47, BUSY=48, RST=8) |
| Sensors | AHTx0 + BME680 on I2C (SDA=9 / SCL=40) |
| Power | VEXT_EN=14, VBAT on GPIO1 |
| Button | BOOT (GPIO0): short press toggles backlight; long press (≥1s) opens power menu |

### Dual SPI (Meshtastic-style)

LoRa and TFT are on **different GPIO groups**, so this port uses two SPI hosts (same approach as Meshtastic `wiscore_rak3312`):

| Peripheral | Bus | Pins |
|------------|-----|------|
| SX1262 | HSPI (SPI3) | 5 / 3 / 6 |
| ST7789 | FSPI via LovyanGFX | 13 / 11 / 12, DC=42, BL=41 |

`radio_init()` calls `SPI.begin()` on the TFT pins first; LovyanGFX uses `bus_shared=true` to avoid a second `spi_bus_initialize` hang.

## Build

```bash
pio run -e rak_tap_v2_companion_radio_touch


B0N3-GD:
Thanks for the port, the board glue itself looks solid. Three things need sorting before this can merge, all about keeping the shared code clean for the other five boards:

1. src/main.cpp picks up bring-up scaffolding that would run on every board: the delay(1500) at boot, the PSRAM probe that halts on failure, and the Serial.flush() calls in the boot path. Please remove them or gate them behind the RAK board define.

2. The LVGL draw-buffer bump from 240 to 320 lines applies to all boards as written. Gate it to the Tap build so the existing S3 boards keep their current RAM footprint.

3. The home-screen chart width block changed for all boards. Keep the original shared behavior and put the Tap-specific sizing behind your board define.

With those three addressed I am happy to merge this. If you would rather I make the changes on top of your branch, say the word and I will push them with your authorship kept intact.

Ethan-chen1234-zy:
Pushed a fix for all three points — shared paths are back to upstream behaviour on the other boards, with RAK-only logic behind HAS_RAK_TAP_V2:

main.cpp — bring-up scaffolding (delay(1500), PSRAM probe, extra Serial.flush() / boot logs) is gated to RAK only; other boards use delay(200) and the original boot path again.

LVGL draw buffer — 320px band width is RAK-only; other S3 boards stay at 240px.

Home chart width — restored the T-Deck/Tanmatsu #if guard and added HAS_RAK_TAP_V2 to that group for landscape RSTRIP trimming.

Verified locally: rak_tap_v2_companion_radio_touch and LilyGo_TDeck_companion_radio_touch both build.

Let me know if anything else needs adjusting — happy to squash or tweak further.

B0N3-GD:
Merged, thank you for the quick turnaround on the review items. Two small things I fixed on your branch before merging, for future reference: a stray closing paren had landed inside the CAP_LARGE_SCREEN block (that arm only compiles on the Tanmatsu target, so S3 builds could not catch it), and the power-menu card height in the shared else-arm drifted from 206 to a computed 196. All three shipping targets build green with the port in. Welcome aboard, the Tap V2 will ship with the next test build.

PR #138, opened 2026-07-08
m9 stable

446564:
not complete, see port doc.

mostly UI quirks now, all hardware etc working.

NOTE: there is a bug fix included for battery logging to sd when the sd is not initialized yet, i.e. the user has not opted in and so the directories do not yet exist.


B0N3-GD:
Merged, thank you. This is a big step for the M9: hardware bring-up complete, and several of these fixes help the other boards too. The un-gated navMarkDirty in setupShowStep (with the LVGL address-reuse explanation) and the lockscreenReveal paint-before-backlight fix are both genuinely shared improvements, and the SD battery-logging guard closes a real hole.

Three small things I fixed on the branch before merging: the control-center sizing lost its '#elif HAS_TDECK_GT911' line (which folded the 200 px card into the Tanmatsu arm and dropped the T-Deck into the portrait else), an unreachable duplicate block after the M9_KEY_HOME return, and a typo in the lock hint. Also resolved the merge with the RAK Tap V2 port that landed this morning (the landscape home-chart gate is now the union of all four boards).

All four targets build green (T-Deck, Heltec V4, Tanmatsu, M9). Ships in the next test build.

PR #179, opened 2026-07-24
touch: SD failure recovery, chat-store diagnostics, and a segmented message store

Vybo:
## What

Three related layers of work on the touch-UI storage stack, driven by field debugging on a T-Deck:

1. **SD failure recovery** — a wedged/removed card is now detected at runtime and remounted; it used to play dead until reboot.
2. **Chat-store health diagnostics** — save failures were 100 % silent on a serial-less device; now they're visible, classified, and self-healing.
3. **Segmented message store** — chat history persistence rewritten from one wholesale-rewritten file to append-friendly segment files.

## Layer 1: SD wedge recovery

Root cause chain (verified against framework-arduinoespressif32 2.0.17 sources): a runtime SD failure is a one-way door — a command timeout latches `STA_NOINIT` (recoverable only by a re-init handshake), while a data-phase error leaves every later op failing `FR_DISK_ERR` with no re-init ever attempted. Meanwhile `SD.cardType()` returns the type **cached at mount** (only `SD.end()` resets it), so every presence gate kept passing, `s_sd_mounted` stayed true, and the file-manager removal poll compared against a value that could never occur. Nothing re-ran the mount ladder until reboot (which power-cycles the card via the shared GPIO10 rail).

The fix: any SD user on any task stamps a volatile flag on failed I/O; `sdHealthTick` (loop task) verifies with real I/O (root-dir open → an actual SEND_STATUS transaction; a recoverable card self-heals right in the probe), drops the mount and re-runs the ladder when the card is dead, and runs a reinsert watch while unmounted (a reinserted card used to *look* recovered while chat history kept writing into the unregistered VFS). The FM poll now probes with real I/O — physical card removal is detected for the first time. All unmount paths hold off under an in-flight history write (`SD.end()` under an open handle = `FR_INVALID_OBJECT`/EBADF forever). Background probing pauses in the sleep regime. `max_files` raised 3 → 6 (telemetry rewrites alone hold 2 handles; the history flush lost the open() race with ENFILE while the card was fine).

Field-confirmed root cause of the "autosave fails, reboot-save works" mystery: the flush rode the **core-0 tile worker — the Wi-Fi core** — where prio-23 driver tasks starve a prio-1 task for hundreds of ms and `sd_diskio`'s busy-waits measure wall time: healthy writes surfaced as EIO. History flushing now runs on a dedicated core-1 task, with a configurable sync-fallback to the loop task after N consecutive failures (Settings → General, default 2; TouchCfg v40).

## Layer 2: diagnostics

- About page → **Chat store** panel: resolved backend (exposes the boot-with-unmountable-card trap where history silently goes to internal flash), segment count + byte total (writer-maintained — zero file I/O at 1 Hz), last save result.
- Failure stage + errno in plain text (`open failed: too many open files (e23)`) on the About panel, the repeated-failure alert, and the reboot-time last-chance-save alert. This taxonomy is what turned the field reports into diagnoses.
- Reboot-time persist hardened: a worker stalled past the 9 s wait counts as unwritten data; a failed last-chance save warns before the reset instead of rebooting into silent loss.

## Layer 3: segmented message store

History persists as `<data root>/msgs/seg_<first_seq>.bin` (256 records each; per-record monotonic seq; real subdirectory on FAT backends, legal flat names on SPIFFS):

- **Appends are ~240 B** to the active segment. Flush coalesce drops 30 s → 2 s (SD) / 5 s (SPIFFS): the hard-cut loss window shrinks ~15×, and the SPI bus-collision window shrinks from a 2 s streaming rewrite to milliseconds.
- **Deletes compact one segment**, rewritten from the RAM ring (disk holds exactly the ring's record set — no read pass); files shrink, fully-aged segments unlink. The old store never shrank: deleting 1000 messages left the file at its high-water size forever.
- **Corruption quarantines one segment** instead of discarding the entire history; crash-ragged tails are trimmed at the record boundary; `File::size()` is never trusted (the Tanmatsu FFat metadata layer lies).
- **Migration is verify-then-delete** with full rollback: old formats (split v6 file, legacy combined) still load; segments are written, read back, and verified before the old file is deleted; a failed migration leaves the old file authoritative and retries.
- The RAM ring and all UI consumers are **unchanged** (slots stay live handles) — this is persistence-only.
- The scheduler moves one bounded job at a time (append batch or one-segment compact) through the existing busy/req worker protocol; the loop task owns the segment table and commits results. Card remounts/formats rebuild the table from RAM and re-land history one segment at a time — no more multi-second UI freezes on recovery.
- Two adversarial review passes over the refactor surfaced 18 bugs, all on crash/failure paths (the on-device happy path never touches them) — fixed in the last two commits, each documented in its message. Headline classes: a failed worker job being committed as durable; migration not being crash-atomic (a power cut left a partial segment prefix shadowing the intact old file — now gated behind a post-verification commit marker); migration assuming a linear ring on its mid-session retry (now walks chronologically and rejects non-monotonic seqs); partially-failed appends duplicating or misaligning records (segments now carry an "untrusted content" flag whose repair is a full rewrite that runs before any append); resync deleting durable files up front (now replaces in place and sweeps leftovers only after the re-land); and the sync drain touching files under a stalled worker's open handle (refused outright — with FF_FS_LOCK=0 that can cross-link clusters). The loader additionally applies a strictly-monotonic seq filter, making it idempotent against anything an interrupted write leaves behind.

## Lifecycle fixes found along the way

- In-session FAT32 format left `/meshcomod` uncreated → every chat/telemetry write failed with ENOENT until reboot; the format path now recreates the data dirs and resyncs the store.
- `sdRestoreApply` ("Copy internal data to SD") rebooted without persisting — dropped everything since the last flush.
- The boot SPIFFS→SD adoption now creates the segment dir (flat SPIFFS names would otherwise be silently skipped).
- Deleted messages truly leave the disk (tombstone-skipping writers), and a legitimately empty store loads instead of being quarantined as corrupt.

## Compatibility

**One-way format**: firmware older than this PR cannot read segment files, and the old single file is deleted after verified migration. Rolling back firmware means losing chat history (identity/contacts/settings unaffected).

## Home-screen chip

The Commander legend row (above the TX/RX chart) carries a compact store chip on its right half: segment count + total bytes normally, `migrating` in amber while an old-format store hasn't converted, `FAIL xN` in red while saves are failing. The full diagnosis stays on the About page.

## Testing

Every commit built green on all five SD-relevant touch envs (T-Deck, V4 TFT, V4-R8, pager SX1262, ThinkNode M9). Field-tested iteratively on a T-Deck through the whole diagnosis arc (wedge → remount → EIO/EBADF/ENFILE classification → core-0 starvation proof). Segmented-store on-device matrix (migration boot, hard-cut loss window, delete shrink, card swap resync, format resync) in progress.


Vybo:
The mechanism works, but I still have trouble with SD I/O after a few hours of uptime, so there's some deeper issue with SD handling that I need to solve first.

B0N3-GD:
Landed on `main` (merge `33306d9`) — thank you for this, it's a genuinely deep piece of work on the SD stack.

While bringing it in we tracked down the **"SD I/O fails after a few hours"** issue you flagged. Root cause: at the segment-table cap (`s_seg_count == k_ui_seg_max`), `segBuildJob` emitted a create-append that `segCommitJob` couldn't record (it only tables a create while `s_seg_count < cap`). So the file was written but never tabled, the durability watermark froze, and the same append re-armed every cycle — a permanent write-spin that leaked orphan `seg_*.bin` files as the ring shifted the target `first_seq`, until the card filled and every SD write failed. It only shows after hours because `s_seg_count` creeps from its ~20 steady state (5000/256) up to the 24 cap via delete fragmentation + step-1 appends starving step-2 retirement.

Fix (the only change on top of your branch): at the cap, defer the create-append instead of emitting an untableable one — skip step 1 (before the watermark advance, so it's never moved past un-written records) and fall through to step 2, which retires the oldest segment first. At the cap that segment has already aged out of the ring (24×256 > 5000), so it unlinks and frees a slot with no live-record loss; the deferred records are the newest and land next cycle. `saveMsgsToStorage`'s early "durable" return is guarded with `segMoreWorkPending` so a deferred append reports a failure rather than a false success.

Full credit is yours — your commits are preserved in `main`'s history, and this closes rather than merges only because the fix couldn't ride your fork branch. Builds green across the whole board matrix (8 S3 + P4 + Tanmatsu).

One honest note: the cap fix is reasoned + compiles, but it still deserves a multi-hour on-device soak against your original repro before we'd promote it to stable — if you get a chance to run it, that confirmation would be gold. Thanks again.

PR #196, opened 2026-07-31
fix(esp32): persist preferences safely off the UI task

### LINKED ISSUE #195: Touch settings reset in-session when touch.kv is rewritten
## Summary

Touch UI settings can reset while the device remains up. The first confirmed symptom was **Settings → Sound → Message sound** turning itself back on; the same device has also shown the first-run setup wizard again after a later reboot.

This is not a notification-routing or reboot/deep-sleep issue: the switch itself returns to on during the same uptime session.

## Environment

- LilyGo T-Deck
- reproduced on beta_50 and current beta_53 (`1863ebd`)
- currently using SD-backed storage; a significantly older card-less build showed a similar symptom

## Confirmed behavior

1. Turn **Message sound** off.
2. Close and reopen Sound settings: it remains off, so the toggle write succeeded.
3. Reboot from the UI and reopen Sound settings: it remains off, so boot loading succeeded.
4. Leave the device running on an active mesh: later, without a reboot, the switch is on again and ordinary message chimes play.
5. DM and @mention routing were excluded: the Message switch visibly changed state.

The observed interval is variable rather than a fixed 15 minutes.

## Root cause

`SdNvsPrefs` stores the entire namespace in one flat `touch.kv` file. Every setter currently:

- opens the live file with `FILE_WRITE`, truncating it in place;
- rewrites every key, including unrelated settings;
- ignores short writes and file-open failures (`sdSave()` returns `void`);
- reports integer/blob puts as successful regardless of whether the file was saved;
- clears the in-memory record list before reloading, so a failed/partial load becomes defaults.

This path is exercised in the background at variable times. In particular, a Discovered-ring change is coalesced for 20 seconds and then writes up to two ~2 KB blob chunks; each chunk rewrites all of `touch.kv`. The clock-floor save is another whole-file rewrite every 15 minutes.

When a rewrite is interrupted or short, later records disappear. Missing `snd_msg` defaults to `true`; missing `setup_ok` defaults to `false`, explaining both the sound reset and the setup wizard reappearing.

Possibly related to the partial-settings/setup symptom reported in #194, though that issue is primarily about slow boot and SD health.

## Proposed fix

- Write `<namespace>.kv.tmp`, verify every write and the completed size, then swap it over the live file.
- Keep a recoverable backup during the swap so a reset between renames always leaves one complete copy.
- Recover interrupted `.tmp`/`.bak` swaps on load.
- Preserve the last-known-good in-memory records when storage cannot be read; do not turn a transient I/O failure into default settings.
- Propagate failed file saves from `put*()` instead of reporting success.

Moving the high-churn Discovered ring to its own file can be considered separately; atomic namespace saves should make the current layout safe first.

PixPMusic:
## Summary

- keep file-backed preference namespaces in a shared RAM mirror and coalesce updates for 750 ms
- commit snapshots from a core-0 worker so slow SPIFFS or SD writes never block LVGL callbacks
- alternate between `.kv` and `.alt` snapshots with generation numbers plus payload/footer CRCs
- write the validation footer last, preserving the previous valid slot across resets or short writes
- move discovered-node blobs into their own namespace so advert churn does not rewrite critical settings
- force pending snapshots before reboot, download mode, deep sleep, shutdown, or factory reset

## Root cause

`SdNvsPrefs::sdSave()` synchronously truncated and rewrote the entire namespace for every setter. A transient failure could leave a valid prefix which `sdLoad()` accepted after clearing its RAM state, making missing keys fall back to defaults. The attempted temporary/backup repair made every preference update perform multiple filesystem metadata operations and caused multi-second UI hitches.

The replacement never clears a loaded namespace after a transient storage failure. Setters mutate RAM, and the inactive snapshot is written and verified off-thread. Boot selects the highest valid generation; a torn inactive slot is ignored.

## Storage and migration

The released `.kv` path doubles as slot A and `.alt` is slot B. The first migration writes `.alt` without touching the released file; the next commit rotates back onto `.kv`. There are only two namespace files during migration and steady state—no `.tmp`, `.bak`, rename, copy, or delete cycle.

The key/value payload remains at the start of `.kv`, with the new footer appended after it, so older firmware can still read its records. Snapshot payloads are capped at 64 KiB to bound RAM and storage use on smaller boards.

Discovered nodes migrate into `discovered.kv` / `discovered.alt`. Legacy chunks are removed from the touch namespace only after the separate snapshot has survived a reboot.

## Verification

- `platformio run -e LilyGo_TDeck_companion_radio_touch`
- `platformio run -e heltec_v4_tft_companion_radio_usb_tcp_touch`
- `git diff --check`
- app-only flash to a TDeck with NVS, SPIFFS, SD, partition table, and user data preserved
- hardware migration: `.kv` -> `.alt` verified in 1129 ms on the worker
- hardware reboot: `.alt` loaded and rotated back to `.kv` in 1003 ms on the worker
- exact commit `2001ef7` flashed and verified; subsequent background snapshot completed in 753 ms
- no preference-write latency appeared in the UI stall log; observed stalls remained in the pre-existing GPS/LVGL paths

Fixes #195

## Checklist

- [x] no synchronous preference writes on normal setters
- [x] bounded two-slot storage
- [x] released `.kv` migration without a wipe
- [x] TDeck and Heltec V4 builds pass
- [x] TDeck hardware migration/reboot tested
- [x] DCO signed-off commit

### DIFF
--- a/src/helpers/esp32/SdNvsPrefs.cpp
+++ b/src/helpers/esp32/SdNvsPrefs.cpp
@@ -4,6 +4,7 @@
 #include <FS.h>
 #include <SD.h>
 #include <SPIFFS.h>
+#include <errno.h>
 #include <string.h>
 
 // ----------------------------- backend selection -----------------------------
@@ -59,23 +60,45 @@ bool SdNvsPrefs::useNvs() {
 // ----------------------------- begin / end -----------------------------
 bool SdNvsPrefs::begin(const char* ns, bool readOnly) {
   _read_only = readOnly;
-  _sd.clear();
-  _sd_loaded = false;
   if (_nvs_open) { _nvs.end(); _nvs_open = false; }
 
   if (fileMode()) {
-    snprintf(_path, sizeof(_path), "%s/%s.kv", s_file_dir, ns);
-    sdLoad();
-    _sd_loaded = true;
+    char path[sizeof(_path)];
+    snprintf(path, sizeof(path), "%s/%s.kv", s_file_dir, ns);
+    if (strncmp(_path, path, sizeof(_path)) != 0) {
+      strncpy(_path, path, sizeof(_path) - 1);
+      _path[sizeof(_path) - 1] = '\0';
+      _sd.clear();
+      _sd_loaded = false;
+    }
+    const bool had_cache = _sd_loaded;
+    const bool loaded = sdLoad();
+    if (loaded) _sd_loaded = true;
     if (nvsHasLegacyData()) _nvs_open = _nvs.begin(ns, true);   // RO migration source
-    return true;
+    // A transient storage outage must not turn a previously-loaded namespace
+    // into an empty one. Readers can keep using the last-known-good mirror;
+    // writers fail so callers do not report a save that never reached disk.
+    return loaded || (readOnly && had_cache);
   }
 
-  if (useNvs()) { _nvs_open = _nvs.begin(ns, readOnly); return _nvs_open; }
-  snprintf(_path, sizeof(_path), "/meshcomod/%s.kv", ns);
-  sdLoad();
-  _sd_loaded = true;
-  return true;
+  if (useNvs()) {
+    _sd.clear();
+    _sd_loaded = false;
+    _nvs_open = _nvs.begin(ns, readOnly);
+    return _nvs_open;
+  }
+  char path[sizeof(_path)];
+  snprintf(path, sizeof(path), "/meshcomod/%s.kv", ns);
+  if (strncmp(_path, path, sizeof(_path)) != 0) {
+    strncpy(_path, path, sizeof(_path) - 1);
+    _path[sizeof(_path) - 1] = '\0';
+    _sd.clear();
+    _sd_loaded = false;
+  }
+  const bool had_cache = _sd_loaded;
+  const bool loaded = sdLoad();
+  if (loaded) _sd_loaded = true;
+  return loaded || (readOnly && had_cache);
 }
 
 void SdNvsPrefs::end() {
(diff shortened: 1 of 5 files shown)


PixPMusic:
Hardware test update from a T-Deck flashed with this PR (`1c761e7`), without wiping its existing data:

- Baseline at 10:17 PDT: `/prefs/touch.kv` was 4,202 bytes, parsed exactly to EOF as 8 valid records, with `setup_ok=1`, `snd_dm=0`, `snd_msg=0`, `snd_men=1`, and `snd_vol=10`.
- Rechecked after 1 hour 3 minutes: all five values were unchanged. The file was still 4,202 bytes / 8 valid records, with no `.tmp` or `.bak` leftovers.
- Rechecked again after 3 hours 55 minutes total: all five values were still unchanged, and the file remained structurally valid with no transaction leftovers.
- Preferences stayed on internal SPIFFS while the SD card remained mounted as secondary storage.
- The whole-file hash changed between snapshots, as expected: both discovered-node chunks contained live updates, and the only decoded `cfg` change was the advancing `clock_floor`. Both `disc`/`disc1` chunks remained valid version-1 blobs with 24 records each.

This covers roughly 15 potential 15-minute rewrite intervals without reproducing the message-notification setting reverting.

PixPMusic:
Hardware follow-up on the T-Deck:

- Confirmed the release workflow uses PlatformIO 6.1.19 in release mode (`-Os`); there is no CI-only optimization.
- A clean local build of beta_54 (`d52fcc47`) remained responsive. After first-time screen construction, observed LVGL redraw stalls were about 281–397 ms.
- A clean build of this PR reproduced the regression immediately: `ui:lvgl` stalls of 3459, 3201, 4360, 3012, and 3214 ms while switching tabs.
- Root cause: `onLvTabChanged()` synchronously persisted `meshTouch/tab` inside the LVGL event handler. The atomic temp/verify/rename sequence made that dead one-byte write take several seconds on SPIFFS. Nothing reads the saved `tab` key.
- `540084a` removes that unused write. Flashed app-only without wiping; preferences survived, and hardware testing confirms tab switching is “way better.”

There is still a separate background stall to address: one `ui:disc 11165ms` sample. `saveDiscovered()` writes `disc` and `disc1`, causing two atomic `touch.kv` rewrites. That is not responsible for the per-tab delay, but it should be optimized separately.

PixPMusic:
Rewrite is now on the branch as `2001ef7`.

The TDeck was flashed app-only with user data preserved. The released `.kv` migrated to `.alt`, survived a reboot, and then rotated back to `.kv` as intended. The physical SPIFFS commits still take roughly 0.75–1.13 s, but they now run on the worker: none showed up as a preference-related UI stall, unlike the prior ~3 s callback hitch.

The device is currently running the exact committed image. The useful hands-on check now is changing a few settings while switching tabs; the serial stall log is ready if anything still leaks onto LVGL.

B0N3-GD:
Merged, thanks. Solid analysis and a careful fix, the layout constraint in that header was handled exactly right. Ships in beta_55 with credit in the notes. One ask for the notes we already know about: after running this build, downgrading reads the plain .kv which can be one snapshot behind. If you get the chance to run it on a Tanmatsu that would close the last untested platform.

PR #197, opened 2026-08-01
feat(touch): add T-Pager UI size presets

PixPMusic:
## Summary

- expose Small, Medium, and Large UI-size presets on T-Pager
- keep map and status chrome compact while scaling semantic text
- bound short-screen chat rows, contacts, message menus, and message-info content
- page-scroll oversized focused content with the Pager encoder before moving focus

## Why

The existing T-Pager text is difficult to read, but applying the large-screen global geometry scale makes the 480x222 viewport unusable. Content and controls run off-screen, tall messages are skipped by focus navigation, and ACK details can become unreachable.

## Behavior

- Small preserves the current 12/14/16 px typography and is the default
- Medium maps those roles to 16/18/20 px
- Large maps them to 18/20/24 px
- the selected preset is persisted and applies after restart
- existing Pager installs migrate the previously ignored ui_scale value to Small once
- Tanmatsu and other boards retain their existing mapping and behavior
- colors and contrast are unchanged

## Verification

- uvx --from platformio platformio run -e tlora_pager_lr1121_companion_radio_touch
- uvx --from platformio platformio run -e tlora_pager_sx1262_companion_radio_touch
- uvx --from platformio platformio run -e LilyGo_TDeck_companion_radio_touch
- git diff --check

All three builds pass. Hardware layout validation is still in progress, so this PR is intentionally a draft.

## Model disclosure

Model: GPT-5.6 Sol | Harness: Codex in T3 Code

PR #198, opened 2026-08-01
fix(touch): keep enlarged Pager UI readable

PixPMusic:
## Summary

Follow-up to #197 after the initial UI-size preset change was merged during hardware validation.

- **_rebuild every checked-in touch font blob from OFL-licensed Montserrat and Noto sources instead of Arial Unicode MS_**
- derive T-Pager Home and composer geometry from the selected font line heights
- widen the Home launcher strip for Medium and Large labels
- add Pager-only 20/24 px multilingual fallbacks for accented Latin, Greek, Cyrillic, Arabic, punctuation, and units
- pin the source versions, archive and font hashes, glyph ranges, and `lv_font_conv` 1.5.3 in a reproducible generator
- include the font copyright notices, full SIL OFL 1.1 text, and maintainer regeneration commands
- widen the Pager Contact Details action sheet to match the sibling chat-details layout, with compact fixed-size actions and overflow scrolling when needed
- widen the Pager Contacts More menu so `Auto-add settings` and longer options do not clip at larger text presets
- size the Mentions header and two-line message cards from the active font metrics instead of fixed-height boxes
- make Pager wheel-hold and Backspace-hold close any shared AppPage, including Store and all Lua apps
- propagate the taller fallback line boxes into LVGL layout so mixed-script and multiline input cannot clip
- size Pager chat rows from the two live text roles while keeping time/unread metadata compact
- use one composer-width calculation for construction, rotation, and keyboard teardown
- preserve Small geometry and all non-Pager behavior
- leave colors and contrast unchanged; the reference photos were taken in a dark room

## Why

The initial presets made semantic text readable, but Medium and Large could still overlap fixed Home/composer geometry and non-ASCII glyphs fell back to visibly smaller fonts.

The previously checked-in fallback blobs, including the original sizes, were generated from Arial Unicode MS. This replaces the complete `extras_font_*`, `extras_lat_*`, and `star_font_*` set with redistributable OFL sources and records exactly how to reproduce it.

## Verification

- Pager LR1121: PASS — RAM 30.5%, flash 83.5%
- Pager SX1262: PASS — RAM 30.5%, flash 83.5%
- T-Deck regression build: PASS — RAM 30.7%, flash 76.6%
- Heltec V4 regression build: PASS — RAM 4.5%, flash 86.8%
- Tanmatsu-only 20/24/28 px assets: PASS — generated deterministically and syntax-checked with `HAS_TANMATSU` (the full ESP-IDF checkout was not present locally)
- `scripts/build/gen-touch-fonts.sh --check`: PASS
- Arial Unicode MS reference scan: PASS — no references remain
- `git diff --check`: PASS
- independent Codex review: PASS after findings were fixed and re-reviewed
- independent Claude Opus 5 xhigh review: PASS after findings were fixed and re-reviewed

Hardware layout validation is still in progress, so this PR is intentionally a draft.

## Model disclosure


PixPMusic:
This is mostly done--I think the big one still to touch is the popup menu buttons when you click settings on a channel or similar

But it's mostly there--might make one more "too large" size where just message text is even larger. I'm blind as a bat and I would love for it to be obnoxiously large but still scroll-able. Getting there.

Might also do the work for this on the T-Deck. This would nice to have even though the displays are tiny. Some of us have bad eyes 😭

B0N3-GD:
Reviewed and build tested against current main. It merges clean, Pager flash goes from 84.3 to 87.7 percent with about 487 KB headroom left, RAM grows 16 bytes, and the other boards are unaffected since the font files compile to empty units off the Pager. Happy with it. Holding it to land together with #210 in the next Pager focused beta rather than in beta_57, which was already large. Mark it ready once your hardware validation is done and it goes in.

B0N3-GD:
Reviewed. The geometry work (deriving Home/composer/chat-row sizing from the live font line heights instead of fixed constants) is the right fix for the Medium/Large overlap, and keeping Small and all non-Pager paths untouched keeps the blast radius sensible. Verification table with RAM/flash per target is appreciated.

Two notes before this comes out of draft:

1. main has moved — MeshCore 1.17 landed in beta_60, and today core-v1.17.1 on top of it (contact-table boot crash + an index regression). Please rebase; the flash figures in particular are worth re-measuring, since both Pager envs were already at 87.0% and the 1.17 core plus the two 20/24px fallback fonts here both add.
2. `extras_font_20.c` / `extras_font_24.c` are ~21k generated lines. Please note in the PR body the exact generator invocation (ranges + symbol list) used to produce them, the way the existing extras font documents its ranges — so the next person regenerating them can reproduce these byte-for-byte rather than guessing.

Mark ready once rebased and I'll build the full matrix and merge.

PixPMusic:
@B0N3-GD I was trying to track down that contact index bug but I've been away from home most of the day.

I've got a few more places on the Pager I want to work on for UI consistency as there are some apps you can't even exit right now. Some (snake) have no way I can tell to input. Etc

Day job has been busy but I want to wrap this up by Thursday. Going camping ! Need my mesh radio working!

B0N3-GD:
@PixPMusic stop hunting the contact index bug — it's found and fixed, don't spend your evening on it.

Root cause: 1.17 reserved `MAX_ANON_CONTACTS` (8) slots at the head of `contacts[]`, so real contacts start at index 8. `getNumContacts()` was updated to exclude them, `getContactByIdx()` was not. Every caller pairing the two read 8 empty slots and could never reach the newest 8 real contacts — invisible in the list, and action-sheet operations resolved to a blank slot. Same reserved-slot handling also caused the #249 boot loop, via `resetContacts()` claiming those slots while our lazily-allocated PSRAM table was still NULL.

Both fixed in `core-v1.17.1` (pinned on main now), full matrix green, going out as beta_61. Details in #252 and #249.

Separately, from your note: "apps you can't exit" and Snake having no usable input on the Pager are worth their own issue — that's real and I'd rather it be tracked than lost in a PR thread. File it (or say the word and I will) with the specific apps and I'll take it.

And no pressure on Thursday. Camping with a working radio beats a merged PR — the burn-in on #234 is worth more after beta_61 lands anyway.

PixPMusic:
At this point @B0N3-GD I think this is ready!

PixPMusic:
Major callout: I changed the font for *all* devices for the Extras font for licensing reasons. We shouldn't be distributing a closed font (!!!)

B0N3-GD:
Merging. Thank you for catching the font licensing — that is exactly the kind of thing that quietly becomes a real problem for a GPL project, and you did not just swap the files, you made it auditable: pinned releases, SHA-256 verification, a `--check` mode that proves the checked-in C files match a clean regeneration, and NOTICE + FONT_LICENSES.md documenting all four upstreams. That is the standard this should have been held to from the start.

Verified here before merge:

- **All 8 S3 boards build green** on current main (core-v1.17.4, with today's contact-table and P4 receive fixes).
- **Flash went down, not up**, despite adding the 20/24 px sizes: both Pagers are now **83.5%**, from 87.0% before. Tightest board is heltec_v4_r8 at 90.0%, V4 TFT at 86.8% — worth keeping an eye on the R8, but there is headroom.
- Merges cleanly onto main; you had already rebased through the 1.17.4 work.

One follow-up rather than a blocker: @pisti87 reports on #232 that Hungarian still shows wrong characters on beta_61. If that is missing glyphs, the Latin Extended coverage in the new Noto set likely fixes it — I have asked him to retest once this lands and to say whether he sees empty boxes (font) or wrong letters (text/encoding).

pisti87:
PixPMusic now beta_62 hungary charset is okay. Thank you

Issue #194, opened 2026-07-30
Extremely slow boot


scsirob:
I am running wadamesh via Launcher on a Liligo T-Deck. Boot time is almost three minutes, is this normal?
When Launcher passes control to Wadamesh, the all-white logo comes on and stays there for close to 3 minutes. After that the corners turn pale blue and a few seconds later the main UI shows up. What could cause such long boot times?

This happens on beta40, beta45 and now beta_50

B0N3-GD:
do you an sd card in the tdeck and native install or launcher?

scsirob:
There is a 4G SD card in the T-Deck. I also tried a fresh 32G card, got even longer boot time.
Wadamesh starts via Launcher

B0N3-GD:
okay that is peculiar, in file manager when you click the sd does it open the folder. (even if it doesn't list he size of the sd card)

scsirob:
Yes it does open, but shows 0KB (Hold:Format) and when selected shows as empty. When Launcher presents it to a computer as USB device then there are plenty of files such as tiles, and backups of the configuration. Copying internal data to SD also takes very long. Once completed, the SD card still appears empty in the file manager.
When relaunching Launcher and connecting as USB device, the SD card contains about 20 files in the root directory, 198 items under /tiles, 623 items under/meshcomod and 2 files under /bkp

scsirob:
When I remove the SD card, Wadamesh takes about 20 seconds to boot

scsirob:
I went through a number of steps. Reformatted the SD card, boot took even longer after that. Updated Launcher, from 2.6.x to 2.7.2. Reinstalled beta_50. Still same issue. Saw some other weirdness too, such as an #unknown channel appear out of nothing. Loaded different app (Saitama) from Launcher OTA catalog, that worked well. Then loaded Wadamesh beta 53 through Launcher OTA catalog. Now boot time is back to normal. At first boot setup wizard some settings were retained, others needed to be reset. Backup simply hangs or triggers a watchdog. Copy internal data to SD fails with 'no identity'. 
I think I somehow need to factory reset the device.

B0N3-GD:
[...]

Issue #193, opened 2026-07-30
T-Pager: microSD detected and mounted, browsable in Files app, but maps/messages not saved to SD


AtomAlex12:
Device: LilyGo T-Pager

SD card: 4 GB microSD FAT32 

Description

The microSD card is detected correctly by the device. When the card is inserted, a notification appears confirming that it has been mounted. The card also shows up and can be browsed in the built-in Files app.

However, there is no "Store data on SD (reboot)" option (or any similar setting) under Settings → General to redirect maps/messages storage to the SD card. As a result, maps and messages continue to be saved to internal storage only, with no way to move them to the SD card.

Steps to reproduce

Insert a 4 GB microSD card into the T-Pager.
Observe the "SD card mounted" notification.
Confirm the card is detected and browsable in the Files app.
Go to Settings → General and look for an option to store data on SD.
No such option is present.

Expected behavior

Settings → General should include a "Store data on SD" option (with reboot) that, once enabled, saves maps and messages to the SD card instead of internal storage.

Actual behavior

SD card is detected, mounted (confirmed by notification), and browsable.
No setting exists to store maps/messages on the SD card.
All data is saved to internal storage only.

B0N3-GD:
Fixed in beta_56. The SD support was wired but never switched on for the Pager, so the card mounted and browsed while nothing stored to it. The Pager now has the same Store data on SD option, boot storage selection and migration as the T-Deck. Maps cache to the card too once it is the active store.

PR #206, opened 2026-08-01
fix(tpager): arbitrate shared SPI SD lifecycle

PixPMusic:
## Summary

This fixes the T-Pager microSD boot hang by giving the display, SX1262/LR1121 radio, and SD card one shared-SPI ownership model and one coherent SD/VFS lifecycle.

- reuse the `SPIClass` already owned by `TFT_eSPI` for display, radio, and SD
- park every inactive shared-bus chip select before mounting the card
- adopt the boot-time Arduino `SD` mount instead of tearing it down from the UI runtime
- prevent mount/unmount while history, preferences, tiles, SD-info, or notification workers still use the VFS
- use the XL9555 active-low card-detect input with bounded I2C locking and fail-safe `Unknown` handling
- keep identity, preferences, channels, contacts, and history on the same profile backend selected at boot
- make first-time SPIFFS-to-SD migration atomic, resumable, identity-last, and fail-closed
- enable the audited Pager storage capabilities with `CAP_SD=1` and `CAP_FILESYSTEM=1`
- keep destructive SD formatting unavailable on Pager

This is the lifecycle/arbitration completion of the Pager SD support introduced by #193. The non-Pager slow-card retry and clock-negotiation behavior from #194 remains unchanged.

## Failure mechanism

The failing build booted normally without a card but stopped at the boot logo when a card was inserted. The failure was a mount-lifetime mismatch, not an unsupported slot or a clock that needed more retries:

1. `main.cpp` mounted the global Arduino `SD` object during boot and handed that live FAT/VFS to DataStore, preferences, contacts/channels, history, and migration code.
2. `UITask.cpp` independently began with its local mount state unset.
3. Its first storage-health/file-manager path treated the already-mounted object as a cold card and entered the generic retry ladder.
4. That ladder called `SD.end()` before retrying `SD.begin()`.
5. `SD.end()` unregistered the same global VFS while boot consumers still held filesystem references or open files beneath it.

With no card, there was no live SD VFS to invalidate. With a card, the UI runtime destroyed storage lifetime underneath the boot sequence. Delays or additional retries could not make that safe.

The Pager now adopts an existing live mount, never uses the generic teardown/retry ladder, and only changes VFS lifetime after every known consumer is quiescent.

## Shared hardware contract

The T-Pager routes display, LoRa, SD, and NFC over SCK 35 / MOSI 34 / MISO 33 with independent selects:

| Device | Select/control |
| --- | ---: |
| display | GPIO 38 |
| SX1262/LR1121 | GPIO 36 |
| microSD | GPIO 21 |
| NFC | GPIO 39 |
| SD detect | XL9555 channel 10, active-low |

There is one bus owner: `TFT_eSPI::getSPIinstance()`. RadioLib and Arduino SD receive that same object, and each peripheral uses SPI transactions for transfers. VFS lifetime is gated separately because transaction locking alone cannot protect an open `File` from `SD.end()`.

This follows LilyGoLib's Pager sequencing at [`a6cd82f`](https://github.com/Xinyuan-LilyGO/LilyGoLib/blob/a6cd82fae25b0e9c186787ec52f80f832b988a00/src/LilyGo_LoRa_Pager.cpp):

- [`initShareSPIPins()` parks NFC, LoRa, SD, and reset selects high](https://github.com/Xinyuan-LilyGO/LilyGoLib/blob/a6cd82fae25b0e9c186787ec52f80f832b988a00/src/LilyGo_LoRa_Pager.cpp#L180-L195)
- [the shared SPI bus is initialized once before peripheral use](https://github.com/Xinyuan-LilyGO/LilyGoLib/blob/a6cd82fae25b0e9c186787ec52f80f832b988a00/src/LilyGo_LoRa_Pager.cpp#L330-L332)
- [`installSD()` checks SD detect, parks shared pins, and mounts at 4 MHz](https://github.com/Xinyuan-LilyGO/LilyGoLib/blob/a6cd82fae25b0e9c186787ec52f80f832b988a00/src/LilyGo_LoRa_Pager.cpp#L466-L486)
- [`uninstallSD()` and readiness checks serialize access](https://github.com/Xinyuan-LilyGO/LilyGoLib/blob/a6cd82fae25b0e9c186787ec52f80f832b988a00/src/LilyGo_LoRa_Pager.cpp#L489-L503)

Wadamesh has concurrent filesystem owners that the vendor example does not, so this PR adds the VFS-consumer gate on top of that electrical contract.

## Profile and migration safety

An SD card may contain the user's identity, private key, settings, channels, contacts, and message history. The migration rules therefore fail closed:

- an existing SD identity defines an established card profile, including cards created before the new completion marker
- automatic migration only targets an empty profile tree
- an interrupted identity-last migration resumes only when its private owner digest matches the current internal identity
- residual identity-less data without that matching owner marker is never mixed into the current profile
- each file lands through its own temporary file and atomic rename
- identity is copied after all settings/history files, and the completion marker commits last
- every new attempt invalidates an older completion marker before copying
- FAT failure leaves the durable migration breadcrumb armed and keeps the device on internal storage
- chat history follows the same authoritative backend decision as identity and preferences
- manual reconciliation flushes pending history, contacts, discovered nodes, and preferences before copying

No identity, key, or digest contents are logged or included in the diff.

## Capabilities and destructive operations

The Pager's audited storage paths now use the structural flags expected by the UI:

- `CAP_SD=1`
- `CAP_FILESYSTEM=1`

The previous claim that these flags would enable dozens of unaudited Pager features was stale: every filesystem-only site was already Pager-enabled, and the only newly exposed SD behavior is the existing map-source switch.

Known follow-up: the Pager map-source choice currently returns to server/cache mode after reboot. #221 persists that choice and further hardens backend switching; it is no longer required to prevent this PR from removing an existing feature.

SD formatting remains excluded by explicit board guards and is not implied by `CAP_SD`. On the card, Pager factory reset only removes wadamesh-owned `/meshcomod` data and preserves root-level map packs; it waits for active consumers and refuses to erase the internal profile when the card-side wipe cannot be verified.

## Verification

  - success on the capability-enabled #206 head
  - RAM: 101,560 / 327,680 bytes (31.0%)
  - flash: 3,435,745 / 4,063,232 bytes (84.6%)
- affected shared paths also built successfully for:
  - `LilyGo_TDeck_companion_radio_touch`
  - `heltec_v4_r8_tft_companion_radio_usb_tcp_touch`
- `PAGER_SD_DISABLED` is absent
- eight changed files; no display-layout or UI-size changes

## Hardware evidence

Testing used application-only flashes at `0x10000`. No bootloader, partition table, NVS, SPIFFS, coredump, backup, or SD image was written.

Observed on a physical SX1262 T-Pager:

- boot reaches the home UI with and without the existing card inserted
- the card's existing `/meshcomod` profile and chat history are adopted
- display, controls, tabs, Wi-Fi, BLE, and LoRa remain operational while SD is mounted
- Pager-originated public traffic and a real multi-hop ACK were observed with the card mounted
- repeated reset/boot with the card present no longer reproduces the original boot-logo hang
- the final capability-enabled integration build, including #210 and #230, mounted the card-backed preference store, associated Wi-Fi, initialized BLE, reached UI-ready, and ran beyond the prior watchdog interval without resetting

## Review

The implementation received an independent Claude Opus 5 `xhigh` review plus independent Codex review. Reproduced findings around partial migration, mixed-profile recovery, coredump gating, factory-wipe messaging, tile-worker lifetime, history backend selection, contact flushing, cross-core pending counters, stale markers, path truncation, and watchdog-state preservation were addressed and rebuilt.

The final capability delta was independently reviewed against the exact #206 head with no correctness findings. A fresh Claude rerun was intentionally skipped at the user's request.

### DIFF
--- a/src/helpers/esp32/TouchPrefsStore.h
+++ b/src/helpers/esp32/TouchPrefsStore.h
@@ -103,7 +103,8 @@ bool touchPrefsGetUseMiles();
 bool touchPrefsSetUseMiles(bool use_miles);
 
 /** Map tile source: false = tile server + on-device cache (default), true = read tiles off the
- *  microSD card (/tiles/<z>/<x>/<y>.jpg). T-Deck only (the V4 TFT has no SD slot). */
+ *  microSD card (/tiles/<z>/<x>/<y>.jpg). Persistent on T-Deck; the Pager currently resets
+ *  this runtime choice to the server/cache mode at boot. */
 bool touchPrefsGetTilesFromSd();
 bool touchPrefsSetTilesFromSd(bool from_sd);
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -227,7 +227,39 @@ void halt() {
 #if defined(TLORA_PAGER)
 static constexpr const char* kSdMigrationComplete = "/meshcomod/.spiffs-migration-v1";
 static constexpr const char* kSdMigrationCompleteTmp = "/meshcomod/.spiffs-migration-v1.tmp";
-bool meshcomodSdMigrationComplete() { return SD.exists(kSdMigrationComplete); }
+
+// Manual recovery may resume onto an empty card or the same profile, but it
+// must never fill one identity's missing files from another profile. Compare
+// the small identity files without logging their contents before arming the
+// durable migration latch. A read failure is treated as a mismatch.
+bool meshcomodSdProfileMatchesInternal() {
+  static constexpr const char* kInternalIdentity = "/identity/_main.id";
+  static constexpr const char* kSdIdentity = "/meshcomod/identity/_main.id";
+  File internal = SPIFFS.open(kInternalIdentity, FILE_READ);
+  if (!internal) return false;
+  if (!SD.exists(kSdIdentity)) {
+    internal.close();
+    return true;
+  }
+  File card = SD.open(kSdIdentity, FILE_READ);
+  if (!card || internal.size() != card.size()) {
+    internal.close();
+    if (card) card.close();
+    return false;
+  }
+  uint8_t internal_buf[64];
+  uint8_t card_buf[64];
+  bool matches = true;
+  while (matches && internal.available()) {
+    const size_t n = internal.read(internal_buf, sizeof internal_buf);
+    if (n == 0 || card.read(card_buf, n) != n || memcmp(internal_buf, card_buf, n) != 0)
+      matches = false;
+  }
+  if (card.available()) matches = false;
+  internal.close();
+  card.close();
+  return matches;
+}
 #endif
 
 bool meshcomodMigrateSpiffsToSd(bool force) {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -52,7 +52,11 @@ static uint32_t _atoi(const char* sp) {
     #include <SD.h>
     #include <Preferences.h>
     #ifndef PIN_SD_CS
-      #define PIN_SD_CS 39      // T-Deck microSD chip-select (V4-R8 sets 3 in the env)
+      #if defined(TLORA_PAGER)
+        #define PIN_SD_CS PAGER_PIN_SD_CS
+      #else
+        #define PIN_SD_CS 39    // T-Deck microSD chip-select (V4-R8 sets 3 in the env)
+      #endif
     #endif
   #endif
   extern "C" void set_boot_phase(int phase);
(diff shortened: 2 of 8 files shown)


B0N3-GD:
Heads up on a collision before you iterate further: beta_56 (issue #193) flipped the Pager to CAP_SD=1 / CAP_FILESYSTEM=1 and extended the T-Deck-style store-on-SD machinery to it (boot storage selection, migration, SdNvsPrefs routing, plus tloraPagerSharedSPI() exported from the variant for main.cpp's boot mount). Your stack here sets CAP_SD=0 in favour of explicit audited wiring, so a rebase onto current main will meet that head on. No objection to the explicit-wiring approach if it covers the store-on-SD paths too, but as it stands the stack would revert a shipped user-facing feature beyond just the tile source that #221 restores. Happy to talk through which direction wins before you spend rebase effort.

B0N3-GD:
beta_57 is out with the Pager storage still on the CAP_SD=1 path, so the collision from my earlier comment is now shipped behaviour. Whenever you get a moment, let me know how you want to reconcile, happy to help with the rebase either way.

PixPMusic:
@B0N3-GD you basically nailed exactly why this was still a draft being worked on—i wanted to get everything under the CAP_SD and CAP_FILESYSTEM flags.

B0N3-GD (approved):
Reviewed the full stack head (15d33e4) and build-tested tlora_pager_sx1262 + T-Deck + V4 TFT against current main — all three green. The shared-SPI ownership model reads right, and the factory-reset hardening (refusing to erase internal identity unless the SD wipe is verified) closes exactly the failure class we hit in beta_36. Nice work as always.

PR #210, opened 2026-08-02
fix(tpager): stabilize Wi-Fi and BLE coexistence

PixPMusic:
## Summary

This makes Wi-Fi and BLE coexist reliably on the LilyGo T-Pager without treating the ESP32-S3 radios as mutually exclusive and without rebooting to enable Wi-Fi.

- keep Pager LVGL allocations and the synchronous display buffer in PSRAM so controller allocations retain contiguous internal/DMA-capable memory
- initialize saved Wi-Fi before NimBLE at boot, while internal heap is least fragmented
- replace live NimBLE destruction with a bond-preserving handoff: stop advertising, disconnect the phone, associate Wi-Fi, then resume the existing GATT server
- wait for NimBLE's asynchronous disconnect to leave the connection table before WPA starts
- give WPA its full six-second association window after BLE is idle
- restore BLE without rebooting when Wi-Fi initialization, disconnect, or association fails
- disable hidden Wi-Fi auto-reconnect attempts while Pager BLE is resident
- harden the remaining intentional NimBLE teardown paths against the captured FreeRTOS timer race using the repository's pinned dependency-patch mechanism

## Failure mechanism

The T-Pager and T-Deck use the same ESP32-S3; the limitation was firmware lifecycle and timing, not different radio hardware.

The Pager could fail in several related ways:

1. UI/display allocations consumed or fragmented internal DRAM required by the Wi-Fi and NimBLE controllers.
2. Starting a cold radio stack after the full UI was running could fail despite ample PSRAM because controller allocations require contiguous internal memory.
3. WPA authentication could overlap an active BLE connection or a hidden Wi-Fi retry, causing BLE timeouts, watchdog resets, and misleading **Not enough memory** errors.
4. The first handoff implementation destroyed and recreated NimBLE. The six-digit PIN remained the same, but the long-term BLE bond state did not, leaving the phone and Pager with mismatched keys and forcing users to forget and re-pair.

The final boundary is narrower: BLE traffic is quiesced while WPA starts, but the bonded NimBLE host, GATT database, device address, and keys remain resident.

## Runtime behavior

- Wi-Fi and BLE remain enabled and allocated together.
- Enabling Wi-Fi while a phone is connected does not reboot the Pager.
- The active GATT connection closes briefly for WPA, but the bond and BLE preference are preserved.
- WPA starts only after `NimBLEServer::getConnectedCount()` reaches zero.
- A stuck disconnect cancels Wi-Fi after one second and restores BLE.
- Wi-Fi receives a fresh six-second association window after BLE is idle.
- BLE advertising resumes after Wi-Fi connects or the bounded attempt fails.
- A previously bonded phone reconnects without entering the PIN again.
- If the saved AP is unavailable, background WPA retries stop and the Pager remains reachable over BLE.
- T-Deck and other boards retain their existing coexistence behavior.

## Captured NimBLE teardown crash

A Pager coredump also identified a deterministic NimBLE-Arduino 1.4.3 race:

- exception: `InstrFetchProhibited`, `PC=0`, `EXCVADDR=0`
- crashing task: FreeRTOS `Tmr Svc`
- timer callback: NimBLE `os_callout_timer_cb`
- concurrent caller: `loopTask` during `NimBLEDevice::deinit(true)`

NimBLE queued `xTimerDelete()` and immediately invalidated callout state. An already-expired callback could run first and dereference the cleared callback. Routine Wi-Fi recovery no longer tears NimBLE down, but teardown remains valid when Bluetooth is explicitly disabled and during selected OTA cleanup paths.

The project already has a pinned PlatformIO dependency-patch mechanism. This PR uses it for the affected Pager environments and fails closed if the expected NimBLE 1.4.3 source does not match. It does not introduce a library fork or change NimBLE major versions.

## Hardware evidence

Tested on a physical SX1262 T-Pager using application-only flashes at `0x10000`; NVS, SPIFFS, partition data, SD contents, and backups were not modified.

Observed:

- boot reached `[BOOT] ui ready`
- saved Wi-Fi associated before the initial NimBLE allocation
- BLE connected using the phone's existing bond
- enabling Wi-Fi from a BLE-connected state completed without reboot or watchdog
- Wi-Fi associated while NimBLE remained allocated
- BLE resumed communication without forgetting or re-pairing the Pager
- an unavailable saved AP fell back to BLE after the bounded attempt
- the UI remained responsive throughout the handoff

The final disconnect-confirmation, WPA-deadline, and bond-preservation revisions were flashed and exercised on the Pager.

## Verification

  - success; RAM 101,232 / 327,680 bytes (30.9%), flash 3,433,345 / 4,063,232 bytes (84.5%)
- `uvx --from platformio platformio run -e tlora_pager_lr1121_companion_radio_touch`
  - success; RAM 101,168 / 327,680 bytes (30.9%), flash 3,432,313 / 4,063,232 bytes (84.5%)
- `uvx --from platformio platformio run -e LilyGo_TDeck_companion_radio_touch`
  - success; RAM 101,868 / 327,680 bytes (31.1%), flash 3,324,077 / 4,063,232 bytes (81.8%)
- Pager dependency patches are pinned and idempotent
- independent Codex review: clean after the disconnect wait, WPA deadline, allocation-failure fallback, and bond-preservation fixes
- fresh Claude review intentionally skipped at the user's request

## Scope

This PR does not change SD lifecycle, LoRa/SPI arbitration, radio calibration, profile migration, preferences schema, partitions, or UI layout.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -640,7 +640,7 @@ framework = arduino
 board = lilygo-t-lora-pager
 monitor_speed = 115200
 board_build.partitions = variants/lilygo_tlora_pager/partitions_tlora_pager_touch.csv
-extra_scripts = pre:scripts/inject_wifi_env.py, merge-bin.py
+extra_scripts = pre:scripts/inject_wifi_env.py, pre:scripts/build/patch_nimble_freertos_timer.py, merge-bin.py
 
 build_flags =
   -Wno-deprecated-declarations -Wno-unused-parameter -DNDEBUG -DRADIOLIB_STATIC_ONLY=1 -DRADIOLIB_GODMODE=1
@@ -865,7 +865,7 @@ framework = arduino
 board = lilygo-t-lora-pager
 monitor_speed = 115200
 board_build.partitions = variants/lilygo_tlora_pager/partitions_tlora_pager_touch.csv
-extra_scripts = pre:scripts/inject_wifi_env.py, merge-bin.py
+extra_scripts = pre:scripts/inject_wifi_env.py, pre:scripts/build/patch_nimble_freertos_timer.py, merge-bin.py
 
 build_flags =
   -Wno-deprecated-declarations -Wno-unused-parameter -DNDEBUG -DRADIOLIB_STATIC_ONLY=1 -DRADIOLIB_GODMODE=1
--- a/scripts/build/patch_nimble_freertos_timer.py
+++ b/scripts/build/patch_nimble_freertos_timer.py
@@ -1,0 +1,149 @@
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# NimBLE-Arduino 1.4.3 clears a FreeRTOS callout immediately after
+# xTimerDelete(), but xTimerDelete() only queues the deletion. If the timer has
+# already expired, the timer daemon can still run the callback against the
+# cleared callout and call address zero. T-Pager exposes this when Wi-Fi link
+# recovery tears NimBLE down dynamically.
+#
+# Keep the host timer alive until nimble_port_stop() has finished, drain every
+# callout deletion before clearing its callback state, and keep the host event
+# queue alive until those callouts are gone. The barrier is FIFO on the same
+# timer command queue, so any already-due callback runs while both its callout
+# and destination event queue are still valid.
+#
+# Idempotent and deliberately scoped to environments that list this script.
+# All replacements are exact and fail closed if the pinned dependency drifts.
+Import("env")
+import os
+
+LIB_ROOT = os.path.join(env.subst("$PROJECT_LIBDEPS_DIR"), env.subst("$PIOENV"),
+                        "NimBLE-Arduino", "src", "nimble")
+
+HOST_PATH = os.path.join(LIB_ROOT, "nimble", "host", "src", "ble_hs.c")
+PORT_PATH = os.path.join(LIB_ROOT, "porting", "nimble", "src",
+                         "nimble_port.c")
+NPL_PATH = os.path.join(LIB_ROOT, "porting", "npl", "freertos", "src",
+                        "npl_os_freertos.c")
+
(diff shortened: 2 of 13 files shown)


B0N3-GD:
Went through this in depth. The cross board pieces are genuinely good finds, the LVGL realloc migration into internal RAM, the NimBLE server callback delete on an interior pointer, and the stale header includes on the IDF mains. The Pager Wi-Fi first state machine is a big behavioural change though, and it deserves its own beta rather than riding along in beta_57, which already shipped your storage and TCXO work. Two things before we take it. Mark it ready when you consider it settled. And one design question: with saved credentials and the AP out of range, the Pager boots with no BLE until Wi-Fi connects or gets turned off. The UI explains it, but is that the tradeoff you want to keep, or is a timeout fallback worth having? For what it is worth I merge tested it against current main, clean, and both the Pager and V4 envs build.

PixPMusic:
@B0N3-GD I was trying to work through that timing issue and remove that requirement before marking this as ready for review.

PixPMusic:
I've managed to hammer out getting both Bluetooth and Wi-Fi working at the same time with the caveat that enabling Wi-Fi will temporarily cut your Bluetooth connection.

I definitely prefer this to running out of RAM.

I still want to dig more into the architecture differences between this and the T-Deck considering it has no issue in this department--no way the screen resolution is the culprit...but that's the main lead I have.

@B0N3-GD ready for your review

B0N3-GD (approved):
Reviewed and build-tested against current main (with #206/#221 merged in): Pager, T-Deck, V4 TFT, Tanmatsu and T-Display P4 all build green. Three things stood out, all good: the interior-pointer analysis on NimBLEDevice::deinit(true) deleting our placement-new'd callbacks (and the defaultCallbacks-sentinel dodge), the fail-closed NimBLE timer patch being scoped to Pager envs only, and the quote-include fixes for the stale vendored-header trap — that one has bitten us before on the IDF mains. Merging.

PR #219, opened 2026-08-02
Add PIN-selected kid profiles

mikecarper:
Different PIN codes open different local MeshCore profiles, keeping the parent and kid identities, contacts, channels, settings, and messages separate.

It keeps a 2,000-entry encrypted RX log while locked, then decodes what the selected profile has keys for when its PIN is entered.

Tested on T-Deck and both T-Pager radio builds.

### DIFF
--- a/src/DataStore.cpp
+++ b/src/DataStore.cpp
@@ -40,9 +40,47 @@ DataStore::DataStore(FILESYSTEM& fs, FILESYSTEM& fsExtra, mesh::RTCClock& clock)
 #endif
 
 const char* DataStore::_rp(const char* name) {
-  if (!_root[0]) return name;            // default: filesystem root, zero change
-  snprintf(_rpbuf, sizeof(_rpbuf), "%s%s", _root, name);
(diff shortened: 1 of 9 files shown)


B0N3-GD:
Thanks for this, the engineering is genuinely good, the encrypted RX buffer while locked is a clever touch and profile 0 keeping the historical paths shows you understood our migration history. We are not taking multi-profile support right now though: identity and storage are the areas where we have been hardening for weeks and a second identity root multiplies the matrix we have to keep safe (SD adoption, migrations, the new prefs snapshots). Not comfortable carrying that yet.

For the family use case there is a workable path today on the SD-store boards (T-Deck, V4-R8, Pager since beta_56): with Store data on SD enabled the card carries the whole identity, so two cards are two nodes, swap and reboot. The trade-off is messages received while a card is out are missed, which your encrypted buffer solved more elegantly, credit where due.

Your retry PR #207 is merged and ships in the next beta with an on/off toggle in Radio & Mesh. If you want to revisit profiles later, a smaller first step we would take: profile-aware storage in DataStore alone, behind a compile flag, before any UI.

PR #221, opened 2026-08-02
fix(tpager): persist and safely switch the SD map-tile source

PixPMusic:
## Status / stacking

Stacked directly on #206. Merge order: **#206 -> #221**.

#213 is merged, and the current #206 head (`ba39ee4`) already contains the
current `main` base (`84378c0`). GitHub cannot select the contributor fork's
#206 branch as this PR's base, so #221 still targets `main` and temporarily
shows the #206 commits in its GitHub diff. The actual #221 layer is exactly:

- `src/helpers/esp32/TouchPrefsStore.h`
- `src/ui-touch/UITask.cpp`

## Summary

- persist **Map tiles from microSD** on T-Pager instead of resetting it at boot
- switch safely between the SD tile library and internal cache on card removal/reinsertion
- pause the tile worker at a request boundary, preserving queued downloads across a backend swap
- keep `/maps/osm` and user PNG packs read-only while keeping `/tiles` writable and repairable
- keep online maps working through the internal cache while the requested card is absent
- protect T-Deck, ThinkNode M9, and Heltec V4-R8 format/remount paths with the same VFS-lifetime gate

## Failure mechanism

The Pager exposed the map-source setting, but boot explicitly replaced the saved
value with `false`. A valid SD tile library therefore appeared unavailable after
reboot even though both the preference and files were present.

Card transitions also had a backend-ownership failure. The fetch worker snapshots
a request's paths, but it continues dereferencing the shared `s_tile_fs` backend
while opening, reading, writing, removing, and closing files. Repointing that
pointer while a request is active can cross a file handle between filesystems,
and tearing down the SD VFS can invalidate the worker's live FAT handle.

Waiting for the entire queue to drain is unnecessary and makes hot reinsertion
appear broken under a busy map. This PR separates:

- queued/in-flight count, used for progress and memory gates
- one active backend lease, which is the only state that can own a `File`
- a swap request, which stops the worker at the next request boundary

A lifecycle owner requests the pause, waits only for the active lease, moves or
tears down the backend, and then releases the preserved queue against the new
backend. Removal therefore falls back to internal flash without discarding
downloads; reinsertion returns to SD without waiting for the backlog to finish.

Finally, corrupt-cache recovery previously could not safely distinguish standard
offline packs from downloaded cache files. This change treats `/maps/osm` and
user PNG packs as immutable, while `/tiles/*.jpg` is explicitly repairable. It
rejects truncated JPEG framing and, after any decoder failure, deletes and
requeues only the writable cache file.

## Behavioral specification

- the saved source survives reboot on T-Deck and T-Pager
- OSM pack reads use SD while the card is present
- an internal cache remains usable for online fetches while the card is absent
- backend changes pause at a worker request boundary and preserve queued work
- active SD tile work participates in the SD VFS-lifetime gate
- reinsertion returns to SD at the first safe boundary, without draining the queue
- topo remains on its online cache; SD packs remain OSM-only
- standard `/maps/osm` and user PNG packs are never deleted by cache repair/reload
- corrupt or truncated `/tiles` JPEGs are removed, dedup is cleared, and a fresh fetch can be queued
- format/remount paths never invalidate an active tile `File`
- a no-op storage notification cannot release a pause owned by another lifecycle operation

## Data safety

No card formatting, bulk rewrite, or destructive hardware test was performed for
this PR. Normal online map use may add successfully downloaded missing tiles to
`/tiles`; repair/reload never deletes or replaces files from `/maps/osm` or
user PNG packs.

## Verification

- `git diff --check ba39ee4...94db9ec`
- `tlora_pager_sx1262_companion_radio_touch`: success
  - RAM: 101,576 / 327,680 bytes (31.0%)
  - flash: 3,437,033 / 4,063,232 bytes (84.6%)
- `tlora_pager_lr1121_companion_radio_touch`: success
  - RAM: 101,512 / 327,680 bytes (31.0%)
  - flash: 3,435,993 / 4,063,232 bytes (84.6%)
- `LilyGo_TDeck_companion_radio_touch`: success
  - RAM: 102,204 / 327,680 bytes (31.2%)
  - flash: 3,328,717 / 4,063,232 bytes (81.9%)
- `ThinkNode_M9_companion_radio_touch`: success
  - RAM: 96,876 / 327,680 bytes (29.6%)
  - flash: 3,322,069 / 4,063,232 bytes (81.8%)
- `heltec_v4_r8_tft_companion_radio_usb_tcp_touch`: success
  - RAM: 102,888 / 8,388,608 bytes (1.2%)
  - flash: 3,306,145 / 4,063,232 bytes (81.4%)

### Pager hardware

The SX1262 Pager was app-only flashed from the local integration containing the
behavior-changing handoff in `2b80029`. It booted with the card inserted, adopted
the SD-backed profile, initialized the radio, associated Wi-Fi, enabled BLE in
the ordered handoff, and reached UI ready.

With map downloads active/queued, removing the card fell back to the flash cache;
reinserting it switched the map source back to SD without waiting for the queue
to drain. The map and device remained responsive. The final `94db9ec` change is
a narrow ownership guard for a no-op notification; the exact final tree was
rebuilt on both Pager radio variants and all affected shared-SD targets.

T-Deck, M9, and V4-R8 received compile coverage only in this final pass; their
format paths were not exercised because formatting is destructive.

## Review findings and dispositions

A fresh exact-diff Codex review found three lifecycle issues:

1. physical removal still discarded queued jobs and waited on queued + active work
2. manual source switching could leave the worker pause latched
3. T-Display P4's SD_MMC cache was mislabeled as flash

All three were fixed in `2b80029`. The final review then found that a no-op
storage notification could clear a pause owned by another lifecycle operation;
`94db9ec` limits pause release to a function that actually requested a swap.
The resulting diff was re-reviewed with no remaining concrete finding.

### DIFF
--- a/src/helpers/esp32/TouchPrefsStore.h
+++ b/src/helpers/esp32/TouchPrefsStore.h
@@ -103,8 +103,7 @@ bool touchPrefsGetUseMiles();
 bool touchPrefsSetUseMiles(bool use_miles);
 
 /** Map tile source: false = tile server + on-device cache (default), true = read tiles off the
- *  microSD card (/tiles/<z>/<x>/<y>.jpg). Persistent on T-Deck; the Pager currently resets
- *  this runtime choice to the server/cache mode at boot. */
+ *  microSD card (/tiles/<z>/<x>/<y>.jpg). Supported and persisted on T-Deck and T-Pager. */
 bool touchPrefsGetTilesFromSd();
 bool touchPrefsSetTilesFromSd(bool from_sd);
 
--- a/src/helpers/esp32/TouchPrefsStore.h
+++ b/src/helpers/esp32/TouchPrefsStore.h
@@ -103,7 +103,8 @@ bool touchPrefsGetUseMiles();
 bool touchPrefsSetUseMiles(bool use_miles);
 
 /** Map tile source: false = tile server + on-device cache (default), true = read tiles off the
- *  microSD card (/tiles/<z>/<x>/<y>.jpg). T-Deck only (the V4 TFT has no SD slot). */
+ *  microSD card (/tiles/<z>/<x>/<y>.jpg). Persistent on T-Deck; the Pager currently resets
+ *  this runtime choice to the server/cache mode at boot. */
 bool touchPrefsGetTilesFromSd();
 bool touchPrefsSetTilesFromSd(bool from_sd);
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -227,7 +227,39 @@ void halt() {
 #if defined(TLORA_PAGER)
 static constexpr const char* kSdMigrationComplete = "/meshcomod/.spiffs-migration-v1";
 static constexpr const char* kSdMigrationCompleteTmp = "/meshcomod/.spiffs-migration-v1.tmp";
-bool meshcomodSdMigrationComplete() { return SD.exists(kSdMigrationComplete); }
+
+// Manual recovery may resume onto an empty card or the same profile, but it
+// must never fill one identity's missing files from another profile. Compare
+// the small identity files without logging their contents before arming the
+// durable migration latch. A read failure is treated as a mismatch.
+bool meshcomodSdProfileMatchesInternal() {
+  static constexpr const char* kInternalIdentity = "/identity/_main.id";
+  static constexpr const char* kSdIdentity = "/meshcomod/identity/_main.id";
+  File internal = SPIFFS.open(kInternalIdentity, FILE_READ);
+  if (!internal) return false;
+  if (!SD.exists(kSdIdentity)) {
+    internal.close();
+    return true;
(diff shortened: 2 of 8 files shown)


B0N3-GD:
Parked together with #206 for now since it is stacked on it, and the CAP_SD direction discussed there decides both. The tile source switching itself looks worth having once the base question is settled.

PixPMusic:
@B0N3-GD I'm pretty happy with the SD fixes and the map fixes at this point! Marked ready for review and re-based on the current main.

B0N3-GD (approved):
Reviewed the #221 layer (TouchPrefsStore.h + UITask.cpp) on the stack head and build-tested all three S3 touch envs — green. The remount handoff for queued tiles and the lifecycle-pause ownership look right.

PR #234, opened 2026-08-04
fix(touch): stabilize virtualized chat teardown and render scheduling

### LINKED ISSUE #233: T-Pager crashes while scrolling virtualized chat history
## Summary

The LilyGo T-Pager can panic while scrolling through a chat. Two independent hardware crashes produced the same LVGL animation-list corruption signature.

This is not an SD, LoRa, Wi-Fi, BLE, or watchdog failure. It occurs in `loopTask` while LVGL processes animations after the virtualized chat window deletes and recreates visible message rows.

## Hardware and build

- LilyGo T-Pager
- SX1262 environment: `tlora_pager_sx1262_companion_radio_touch`
- Reproduced on two separate integration builds based on current upstream chat virtualization and Pager navigation behavior
- Reproduced while scrolling inside an open chat

## Reproduction

1. Open a chat containing enough history to require virtualized row reflow.
2. Scroll through the conversation with the Pager encoder.
3. Continue crossing materialized-window boundaries so visible message rows are deleted and recreated.
4. The device eventually panics and reboots.

The failure is intermittent per boundary crossing but repeatable during sustained scrolling.

## Crash evidence

Both captured crashes are `InstrFetchProhibited` exceptions in `loopTask`, reached from the LVGL animation timer.

- One attempted an animation completion callback at invalid address `0x66`.
- The other attempted an animation `get_value_cb` at invalid address `0x20000`.
- No firmware or bundled LVGL call site assigns `get_value_cb`, so a non-null value there proves that the animation descriptor is stale or overwritten.
- The matching frames contain no SD, radio, Wi-Fi, BLE, or watchdog origin.

Raw coredumps are intentionally not attached because they can contain arbitrary RAM, including private identity material. They have been preserved locally with matching firmware ELFs and hashes.

## Root cause

This is an interaction between two existing upstream changes:

- `d964c029`: virtualized chat scrolling synchronously deletes and recreates visible bubble rows.
- `85e8681f` / #162: Pager encoder focus remains on those virtualized rows and navigation is synchronously rebuilt after reflow.

During a reflow, `chatVirtClearBubbleWidgets()` deletes the focused row while it is still a member of the Pager navigation group. LVGL object destruction removes that row style and animation state before group removal. Group removal then auto-refocuses another row and calls the application focus callback while the old row is only partially destroyed. Repeating this across every materialized row causes unsafe focus-style and animation churn and eventually leaves a stale animation descriptor in the LVGL animation list.

## Proposed fix

Before deleting any virtualized Pager chat rows:

1. Capture the focused logical message index or stable non-row focus target.
2. Remove the application focus styling while all referenced objects and child labels are still alive.
3. Detach the navigation group before row destruction, preventing LVGL auto-refocus callbacks during destructors.
4. Delete and recreate the visible row window.
5. Rebuild navigation once and restore focus to the recreated logical row or surviving target.

Do not disable scrolling animations globally and do not patch LVGL for an application lifecycle violation.

## Validation required

- Build `tlora_pager_sx1262_companion_radio_touch`.
- Verify encoder navigation remains within long chats across virtualized boundaries.
- Verify composer, header actions, message actions, and back navigation retain correct focus.
- Exercise all Pager UI-size presets and both compact/full chat rows.
- Run sustained scrolling burn-in with incoming messages and normal SD/radio activity.
- Confirm no new crash report is generated.

PixPMusic:
Closes #233. Related: #251.

## Problem

The T-Pager can panic while rapidly scrolling a virtualized chat across message-window boundaries. The matching dumps all fail in LVGL through an invalid callback:

- the original #233 dumps reached corrupted LVGL animation callbacks;
- the beta_60 dump in #251 reaches `lv_timer_exec -> lv_timer_handler` through callback `0x00020000`;
- a fresh Pager dump from the first revision of this PR reaches the same timer call site through callback `0x00000066`.

The fresh dump matters: it was produced by an integration build that already contained the focus-detach change below. That proves the first diagnosis and fix were incomplete. The failure is still not rooted in SD, LoRa, Wi-Fi, BLE, watchdog expiry, or task-stack exhaustion; it is an invalid LVGL callback reached from `loopTask` while scrolling a chat.

Raw dumps remain local because a core dump can contain arbitrary RAM and private identity material. Matching firmware ELFs, hashes, registers, and symbolized call chains are preserved.

## Application-level lifetime hazards

The chat virtualizer had two independent high-churn lifetime paths:

1. **Focused row teardown.** A materialized Pager row could remain in the LVGL navigation group while the virtualizer synchronously deleted it. LVGL could auto-refocus another object during the partially-complete teardown, re-entering application focus styling against objects being destroyed.
2. **Render timer churn.** Every virtual-window boundary scheduled a one-shot chat render timer. Its callback deleted its own LVGL timer node, and while kinetic scrolling continued it immediately allocated another timer. Fast scrolling repeatedly exercised that delete/recreate path in the same timer-list walk where #251 and the new dump fail.

The second dump does not expose heap contents, so this PR deliberately does not claim which individual freed allocation held the bad callback. It removes both application-owned lifetime hazards without patching LVGL.

## Implementation

- Capture the focused logical message index, or a stable non-row target, before a virtualized reflow.
- Restore application focus styling while referenced objects and descendant labels are still alive.
- Empty the LVGL navigation group before deleting rows, then recreate the visible window, recollect navigation once, and restore focus before the next paint.
- Route empty-thread and low-memory replacements through the same protected teardown.
- Stop pre-deleting rows during chat refreshes; replace stale rows only inside the protected virtualizer pass.
- Allocate the chat render timer once, pause it to cancel or idle it, and resume/reset that same timer for later boundaries. It no longer deletes and recreates its LVGL list node during scrolling.
- Keep the existing deferred `lv_async_call` for tree mutation after display refresh.
- Keep existing touch/encoder behavior. No LVGL or dependency patch is required.

## Validation gate

- [x] Rebased on current `main` after beta_61.
- [x] Build `tlora_pager_sx1262_companion_radio_touch`.
- [x] Build `LilyGo_TDeck_companion_radio_touch` as a second keypad-navigation target.
- [x] Build an integration containing only beta_61, #198, and this PR.
- [x] Flash app only at `0x10000`, verify the on-device digest, and boot through `[BOOT] ui ready`.
- [ ] Confirm encoder focus stays inside long chats across virtualized window boundaries.
- [x] Confirm composer, header actions, message actions, and back navigation retain correct focus.
- [x] Exercise compact and full chat rows across Pager UI-size presets.
- [x] Burn in sustained rapid up/down scrolling with incoming messages.
- [x] Repeat with normal SD-backed history and radio activity.
- [x] Confirm no new crash report is generated.

## Current state

PR head `cbdd6d2` is installed as part of local integration `d403ff3` on the T-Pager and boots cleanly. The PR remains a draft until the on-device scrolling and burn-in checks above pass.

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -4117,6 +4117,29 @@ static lv_obj_t* s_nav_prev_scr = nullptr;
 static uint32_t  s_nav_prev_sig = 0xFFFFFFFFu;
 static void navMarkDirty() { s_nav_dirty = true; }
 
+// A virtualized chat rebuild destroys the currently-materialized row objects.
+// Detach the navigation group while every object (and every descendant label
+// remembered by navInvertText) is still alive. Letting lv_obj_del() remove the
+// focused row itself is unsafe: LVGL has already torn down that row's styles and
+// animations when lv_group_remove_obj() auto-focuses a survivor and calls
+// navFocusCb(), which then tries to restore styles on the half-destructed row.
+// The caller must rebuild the group after recreating the object tree.
+static bool navDetachBeforeTreeMutation() {
+  if (!s_nav_group) return false;
+  if (s_nav_styled) {
+    navUnstyle(s_nav_styled);
+    s_nav_styled = nullptr;
+  } else {
+    navRestoreText();
+  }
+  lv_group_remove_all_objs(s_nav_group);
+  s_nav_first = s_nav_last = nullptr;
+  s_nav_count = 0;
+  for (int i = 0; i < kNavMax; ++i) s_nav_objs[i] = nullptr;
+  navMarkDirty();
+  return true;
+}
+
 // Repopulate the focus group when the visible screen actually changes (screen
 // swap, modal open/close, or a within-screen content rebuild). A 2-level
 // child-count signature avoids reshuffling focus during steady-state navigation.
@@ -31169,18 +31192,16 @@ static void chatVirtFreeOffsets() {
 
 #if defined(TLORA_PAGER)
 // Encoder-nav focus survival across the virtualized re-render. The render
-// deletes and recreates every bubble row — LVGL hands focus of a deleted
-// object to the next surviving group entry (the chat header/composer once
-// every bubble is gone), which is how focus "fell out" of the message list
-// mid-history on this no-touch board. The pending logical index below is the
-// message the encoder is steering toward: chatVirtRenderWindow re-aims focus
-// at the matching recreated row via the existing one-shot s_nav_focus_hint
-// rebuild mechanism, and pagerEncoderChatEdgeScroll both sets it (edge detent
-// = neighbor index) and folds further detents into it while a render is in
-// flight — fast turning otherwise stepped focus out from the transiently
-// wrong focus position before the load landed. -1 = idle. The timestamp
-// expires a stale pending target (see the clamp) so a render that never
-// fires can't wedge encoder nav.
+// deletes and recreates every bubble row, so a row pointer cannot preserve
+// focus across the rebuild. The pending logical index below is the message the
+// encoder is steering toward: chatVirtRenderWindow safely detaches the group
+// before destruction, then re-aims focus at the matching recreated row via the
+// existing one-shot s_nav_focus_hint mechanism. pagerEncoderChatEdgeScroll sets
+// the target (an edge detent selects the neighboring logical index) and folds
+// further detents into it while a render is in flight — fast turning otherwise
+// stepped focus out from the transiently wrong focus position before the load
+// landed. -1 = idle. The timestamp expires a stale pending target (see the
+// clamp) so a render that never fires can't wedge encoder nav.
 static int      s_pager_chat_focus_i  = -1;
 static uint32_t s_pager_chat_focus_ms = 0;
 #endif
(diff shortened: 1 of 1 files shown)


B0N3-GD:
Status from our side, so this isn't waiting on us: the gate here is yours — you asked for T-Pager hardware validation and a sustained scrolling burn-in before this is marked ready, and that's the right call for a crash fix in the LVGL teardown path. The analysis in the description matches what the coredumps in #233 show, and the approach (empty the nav group before deleting rows, restore focus while the objects are still alive) is the correct shape.

One thing that changed under you: main moved to MeshCore 1.17 (beta_60), and today to core-v1.17.1 with two contact-table fixes — a boot crash and an index-space regression that made the newest contacts unreachable. Neither touches the virtualizer, but they do change what a burn-in on beta_60 is measuring, and beta_60 itself boot-loops on devices with no stored contacts. Worth rebasing onto current main and doing the burn-in against beta_61 when it lands — a clean run then really is clean.

Mark it ready when the hardware run is done and I'll build-test across the matrix and merge.

PixPMusic:
@B0N3-GD I have been trying to rebase each morning when I see movement

Got another crash log to analyze as I try and hammer out this crash. It's super intermittent—or, I haven't found a consistent reproduction.

PixPMusic:
There's definitely a bug in here still where messages can still exist behind the composer but I feel like this is stable enough to at least stop things from crashing for users. Worth merging and addressing the other issue as a follow-up.

Issue #228, opened 2026-08-03
The unlock animation still says MESHCOMOD instead of WADAMESH


smhawkes:
The unlock animation says MESHCOMOD instead of WADAMESH.

B0N3-GD:
Good catch. The beta_56 fix changed the text the unlock animation draws, but the MESHCOMOD you are seeing is baked into the default wallpaper image itself, both the embedded render and the copy seeded to storage. Both assets are regenerated with the WADAMESH mark and the fix replaces the old seeded copy automatically on the next boot. Ships in the next beta.

B0N3-GD:
Shipped in beta_58. The wallpaper now carries the WADAMESH mark and the seeded copy replaces itself on first boot after the update.

PR #235, opened 2026-08-05
fix(tpager): update lock-screen branding

PixPMusic:
## Summary

- regenerate the T-Pager's native 480x222 lock-screen wallpaper from the shared WADAMESH RGB565 artwork introduced for #228
- preserve the Pager-specific left-aligned composition and clear right-hand status area
- add a deterministic, standard-library generator with a `--check` mode so the board-specific derivative can be kept in sync

## Why

Issue #228 updated the shared 320x240 wallpaper and seeded JPEG, but the T-Pager bypasses that default and renders its own native-resolution RGB565 asset. That separate derivative still contained the MESHCOMOD wordmark, so beta 58 and later continued to show the old name on this board.

The replacement is derived only from the already-merged WADAMESH pixels and the existing Pager background/layout; it does not introduce newly drawn or model-generated artwork.

## Verification

- `python3 scripts/build/gen-lockscreen-wallpaper-pager.py --check`
- `python3 -m py_compile scripts/build/gen-lockscreen-wallpaper-pager.py`
- `pio run -e tlora_pager_sx1262_companion_radio_touch` — PASS (RAM 30.4%, flash 78.0%)
- decoded the generated RGB565 header at its native 480x222 resolution and visually verified the WADAMESH wordmark, teal lock, and clear right-hand status area
- hardware validation on the T-Pager integration build — PASS; the lock screen displays the WADAMESH artwork correctly

## Model disclosure


B0N3-GD (approved):
Generator --check passes against the committed asset (deterministic, stdlib-only — nice touch) and the Pager env builds green on top of today's merges. The Pager finally stops saying MESHCOMOD.

PR #287, opened 2026-08-20
ThinkNode M9 + Heltec V4-R8: 2026-08 audit passes (nav soft-locks, storage, deep sleep, Spectrum at spec)

cvhviz:
## ThinkNode M9 (`68e3b45`, `1b7d793`)

- **Nav soft-locks fixed**: the 0ea242c regression that killed the hardware Back key in nav mode; SUB_MAP over a display-only Lua app stranding the app unreachable; the Back/HOME ladder redesigned so the Control Center, power menu, and confirm modals close in visual order (the Lua send-permission dialog used to be unanswerable on this touchless board); map pan mode can no longer leak across tab jumps or capture arrows under popups.
- **Storage**: "Store data on SD" honored (boot data-store path had no M9 arm), SD-backed chat history, map tile cache prefers the built-in card, atomic threads-index writes (shared fix — a power-cut mid-write used to wipe the chat list).
- **Gate parity**: terminal RX mirror, fullscreen view titles, wallpaper caption refresh, Enter-submit, "Save update bin to SD" and friends widened from their T-Deck-only gates; accent/@-mention pickers suppressed (touch-only chrome with no key path here).
- **Board API**: deep sleep actually drops the power rails now (display refcount + LEDC pin re-route + RTC/digital pad holds) and powers off the LR1110/GPS; battery reads filter Wi-Fi-blocked ADC2 samples; keyboard backlight cache only latches ACKed I2C writes.
- **Build hardening**: the RadioLib old-LR1110-FW patch fail-closes at link time (no unpatched -706 binary can ship), `ENV_SKIP_GPS_DETECT` + `CORE_DEBUG_LEVEL=0` added to the env.

## Spectrum brought to LR1110 spec (`e56ac11`)

Per-bin standby parked in STDBY_RC and re-paid the ~5 ms DIO3-TCXO startup every bin (~2x sweep time) — bins now park in STDBY_XOSC (raw 0x01; RadioLib's `RADIOLIB_LR11X0_STANDBY_XOSC` define is bugged identical to `STANDBY_RC`). One span-wide CalibImage per open with the mesh ±4 MHz calibration restored on close; sweep clamps corrected to the driver-enforced 150–960 MHz on both driver families with failed retunes skipping the bin; the 0 dBm failed-read sentinel no longer wins the peak-hold; opening Spectrum mid-TX now waits (bounded by the dispatcher's own airtime budget) instead of truncating the packet on the air.

## Heltec V4-R8 (`68e3b45`)

Chat history/battery log follow SD adoption, the reinsert watch no longer unmounts the live data store, tile cache prefers SD, `CAP_ROTATABLE` reboot trap defused, SD CS parked at boot. Full record in `variants/heltec_v4/R8_AUDIT.md`.

## Verification

All four touch envs (M9, T-Deck, V4-R8, pager LR1121) compile clean; both fix rounds passed an independent adversarial code review; M9 and V4-R8 validated on hardware. Remaining hardware-verify items are tracked in `variants/thinknode_m9/M9_PORT.md` ("Audit pass 2" + the Deferred list) and `R8_AUDIT.md`.

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -27,8 +27,9 @@ USB) and the [GitHub releases](https://github.com/ALLFATHER-BV/wadamesh/releases
   microSD for all persistent data. Ships through the Tanmatsu launcher store,
   updates arrive as store updates.
 - **ThinkNode M9**: keyboard plus d-pad navigation (no touch), same feature set
-  as the other boards where the hardware allows. New in beta_38; expect rough
-  edges and report them.
+  as the other boards where the hardware allows. Every key and mode is covered
+  in the [keyboard & d-pad guide](THINKNODE_M9_SHORTCUTS.md). New in beta_38;
+  report anything that feels off.
 - **RAK WisMesh Tap V2**: newest port, touch-driven. The browser-flash path is
   fresh; if the flasher cannot open the serial port, put the board in download
   mode manually and retry, and please report it.
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ per-board status.
 - LilyGo T-Deck / T-Deck Plus — env `LilyGo_TDeck_companion_radio_touch` (stable)
 - Heltec V4 + TFT + CHSC6x touch — env `heltec_v4_tft_companion_radio_usb_tcp_touch` (stable)
 - Tanmatsu (ESP32-P4) — built from `tanmatsu/` (ESP-IDF), ships via the Tanmatsu app store
-- Elecrow ThinkNode M9 — env `ThinkNode_M9_companion_radio_touch` (beta)
+- Elecrow ThinkNode M9 — env `ThinkNode_M9_companion_radio_touch` (beta) — [keyboard & d-pad guide](THINKNODE_M9_SHORTCUTS.md)
 - RAK WisMesh Tap V2 (RAK3312) — env `rak_tap_v2_companion_radio_touch` (beta)
 
 ## Architecture
--- a/THINKNODE_M9_SHORTCUTS.md
+++ b/THINKNODE_M9_SHORTCUTS.md
@@ -1,0 +1,97 @@
+# ThinkNode M9: Keyboard & D-pad Guide
+
+The Elecrow ThinkNode M9 has **no touchscreen** — every screen, every setting,
+and every chat is driven by the d-pad, the dedicated function keys, and the
+QWERTY keyboard. If you're coming from the T-Deck or Heltec V4 (touch-first),
+this page is the five-minute tour that makes the board feel native.
+
+## The inputs
+
+- **D-pad** — four arrows around a centre button (**OK**). Arrows move the
+  on-screen focus highlight; OK activates whatever is focused. **Holding OK**
+  is the board's long-press: it fires the held item's long-press action (a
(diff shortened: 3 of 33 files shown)


cvhviz:
Two more commits pushed to this branch (`16e8831`, `844d8c3`). Summary for review, with the cross-board defaults called out first:

**Defaults that change for every board**
- **Map "Show tile z/x/y" line now defaults OFF** (prefs schema v50, one-time flip of existing installs; the toggle in Map options still brings it back). The `z12  12/2105/1376` readout was developer clutter on the map.
- **Companion app-sync ring is now persisted** (`/synchist` append log + `/synccur` cursors on the hot-data FS, restored in `MyMesh::begin()`). It was RAM-only, so any power cycle lost the replay history the phone app syncs from — invisible on USB-powered boards, fatal on the M9 whose only "off" is the hard slider. No behaviour change other than the sync surviving reboots.

**Unchanged for other boards, new pref**
- **Wi-Fi modem power save is now a persisted toggle** (Wi-Fi settings → "Power save (modem sleep)", `wifi_ps` in the NVS Wi-Fi store, applied live once associated). Default **ON everywhere** (today's behaviour); only the V4-R8 defaults OFF (USB-powered Expansion Kit form factor — lower-latency TCP/app link, steadier association on a weak 2.4 GHz path).

**Heltec V4-R8 only — perf pass** (full record in `variants/heltec_v4/R8_AUDIT.md` → "Perf pass")
- FEM LNA defaults ON (prefs v49, one-time flip): the R8 is a V4.3.1-gen board with the KCT8103L FEM, and its switchable ~17 dB RX LNA shipped bypassed.
- Display SPI 40 → 80 MHz (`LGFX_SPI_WRITE_HZ`; `-D` it to `40000000` to step back if a unit shows tearing).
- Async DMA band flush in `LGFXDisplay` (byte-swap into an internal DMA buffer, one no-convert DMA per band, LVGL released immediately; the last band closes the frame transaction so the shared micro-SD gets the bus between frames as before; sync fallback if the buffer can't be allocated).
- micro-SD operating clock 4 → 20 MHz after the proven mount, read-verified (probe file's head byte-compared after the raise, falls back to 4 MHz). `include/SdFastClock.h`, wired at all three mount sites; a no-op on boards that don't define `SD_SPI_FAST_HZ`.
- R8 env boots at `ESP32_CPU_FREQ=240` (setup() used to run at the base V4 env's 80 MHz until UITask bumped it).
- Settings → About gains a `Perf:` row on the R8 (`CPU 240 · TFT 80 MHz DMA · SD 20 MHz · LNA on`) since that board has no reachable serial console.

**ThinkNode M9 only**
- Spectrum page root gets `NAV_SKIP_FLAG` — nothing on it is actionable, so the d-pad no longer highlights the chart; Back closes it as before.
- The tab bar is bypassed entirely: the accent "glow" indicator was a child of the screen, not the bar, so hiding the zero-height bar had left it glowing over the content's bottom edge.

Cross-compiled R8 / V4 / M9 / T-Deck; flashed and booted on my R8 and M9. Hardware items still worth a second pair of eyes: 80 MHz display bus on other R8 units (shared SCLK with the SD), and whether the LNA-on default desensitises in RF-noisy sites (toggle is right there in Radio & Mesh).

Issue #89, opened 2026-07-04
Tdeck no communication with roomservers


EricMW61:
In v21beta it seemed to work, but later versions not. When connecting roomserver it will show login. After login chatscreen will open, but messages will not reach roomserver. Tdeck will transmit, but it's not relayed. Messages will not being retrieved from roomserver. Looks like its frozen altough a send message will show a V in chatscreen but that's all. There is no possibility to log out neither to log in to the roomserver once it is done. 

With other companions (with other firmware) it is possible to send and retrieve messages to and from roomserver. So the roomserver is doing it's job. But messages send with T-deck will not show up on other companions, because roomserver never received them. And messages send form other companions will not show up on T-deck. So Tdeck is not connected although the Tdeck suggests its connected while it shows the chatscreen.

Please may be it's possible to investigate what is missing and to make a button to disconnect or reset connection to roomserver?

PS also local repeater shown in contacts and when selected a chat window will open.

EricMW61:
beta28 looks the issue is resolved. will check thourougly this weekend. Thank you so much for your efforts. Really appreciated. 
Important note: made backup first and installed beta28 including total erase of Tdeck device.

B0N3-GD:
Thanks for the detailed report Eric, this one led us to the actual root cause.

What happens: after a few pushes that don't get acknowledged (easy to hit on LoRa), the room server quietly stops sending anything to that client. The client is supposed to send periodic keep-alives that recover from this, but the firmware never sent them. So the room ends up logged in but dead in both directions, and there was no way to log out or back in. Exactly what you described.

This is fixed in beta_28, now on the test channel. The device keeps the room session alive in the background, logs back in automatically when the session drops, and there is a new "Log in again" option in the room chat's menu (the gear next to the room name). It also warns you when the room only gave you read-only access, because in that case posts get dropped silently.

To try it: Settings > About > turn on "Get test builds (beta)", then install the update. Or flash beta_28 from wadamesh.com (Standalone install, pick beta_28 in the version dropdown).

One thing to know: if the room server itself reboots, it forgets everyone who isn't an admin. That is a protocol limitation, not something we can fix on our side. In that case do one fresh Join with the room password and everything flows again.

Let me know how it behaves for you on beta_28.

EricMW61:
[...]

PR #292, opened 2026-08-22
ThinkNode M9 compass (QMC6309) + GPS motion for Lua apps, and a GPS Compass store app

cvhviz:
Adds the ThinkNode M9's magnetometer to the firmware, widens the Lua GPS
binding, and ships a **GPS Compass** store app that uses both. Merged up to
beta_68; M9, V4, V4-R8, T-Deck and Pager all build.

## Firmware

**QMC6309 driver** (`variants/thinknode_m9/M9Compass.{h,cpp}`, `HAS_M9_COMPASS`
→ new hardware cap `CAP_COMPASS`). The chip was documented on the M9 but nothing
had ever talked to it. Configured from the datasheet's own register map rather
than SensorLib, whose `setOutputDataRate()` writes the ODR into the OSR bits
(Meshtastic inherits that and only works because it free-runs in continuous
mode): 100 Hz, ±32 G, OSR 8, low-pass 4, soft reset with the mandatory explicit
clear, config read back and re-written once if it did not stick. Overflow
samples are kept, flagged and rate-limit logged rather than dropped, so a board
with a huge bias cannot look like a missing chip.

**Why ±32 G on a 0.5 G planet:** measured on hardware, the M9's on-board
hard-iron bias is ≈ −0.34 / −3.4 / −2.8 G — about 7× Earth's field. Meshtastic's
implausible-looking hardcoded extrema were right. At ±8 G the Y axis would sit
within half a scale of the rail before the user's own environment is added.

**Axis orientation, measured not guessed.** It is documented nowhere, and
Meshtastic's M9 driver declares an offset it marks "must be verified on real
hardware" and never uses. Logging the raw vector flat at four headings 90° apart
gives a hard-iron centre of (−0.340, −3.378) and, after subtracting it, N
(−0.055, +0.310), E (+0.268, −0.018), S (+0.013, −0.275), W (−0.225, −0.016) —
so `atan2(x, y)` reads 350/94/177/266 and counts up clockwise: +Y is the top
edge, +X the left. The app ships that as its default, so a calibrated device
points north with no user setup. Full table in `M9_PORT.md`.

**GPS motion** (`src/helpers/WadaNmeaLocationProvider.h`, `HAS_GPS_MOTION`).
MeshCore's provider parses RMC speed/course and then discards them behind a
private member. Rather than patch a dependency, this is a Wadamesh-owned copy —
identical line for line, including the #89 time-sync rules — plus `motion()`.
Course is withheld below 1 km/h because MicroNMEA parses an empty course field
as 0, which is indistinguishable from north.

**SDK:** `wada.sys.gps()` gains `speed_kmh`/`course` (merged with beta_68's
`fix_time`/`lat_e6`/`lon_e6`) and now returns nil while the user has GPS
switched off, not merely when there is no fix. New `wada.sys.compass()` →
`{x, y, z, ovfl}` in Gauss, sensor frame, uncalibrated, behind `caps().compass`
— deliberately no `heading`: that needs hard-iron offsets and an axis mapping,
both of which belong where they can be tuned and persisted per user.
`wada.ui.colors` gains `panel`; `label:width()` takes an optional alignment.

## Fixes found on hardware

- **Every Lua app opened on a white page** on keypad-nav boards: the app body is
  clickable for touch, so `navCollect` harvested it as a focus target and the
  focus highlight's reverse-video fill painted the whole page. New
  `NAV_PASSTHRU_FLAG` (clickable, never a target, children still collected) on
  the body *and* the root — excluding only the body moved the highlight up one
  level rather than removing it.
- **Map re-open cost ~2.5 s, every visit.** Leaving the tab called
  `freeMapTiles()`, so each open re-read and re-decoded nine JPEGs
  (`[STALL] ui:lvgl 2597ms`, twice, identical). Boards with ≥4 MB PSRAM now keep
  the slots; the 2 MB V4 keeps the old free. Measured after: cold open 2598 ms,
  re-opens **245 ms**.
- **The net worker freed its buffers out from under itself.**
  `luaNetWorkerService()` cleared `s_net_pending` before making the request, so
  an app closing during a round trip freed `s_net_buf`/`s_net_body` while the
  other core was still using them. The flag now spans the request.
- The `http_get` buffer was never freed when an app closed; canvas pixel buffers
  could be GC'd while LVGL still drew from them (`cvGc` freed unconditionally
  despite its comment saying otherwise); `mesh.contacts()` did one RTC I2C read
  *per contact* on an I2C-RTC board; touch coordinates ignored the body's scroll
  offset.
- **The manifest `icon` field now works.** `LUA_APPS.md` promised it from the
  start; every Lua app drew the same generic glyph. Icons are named
  (`gps`, `radio`, `chart`, …), not codepoints, so a submission stays reviewable
  and cannot ship a missing-glyph box.

## Store app — `deploy/apps/gpscompass/1.0`

Dial with graduations and a lubber mark, a key/value panel (FIX with a satellite
meter, LAT, LON, ALT, SPD, TGT), bearing/range/relative-turn to a chosen contact,
hard-iron calibration, and per-row imperial/metric units. Falls back to GPS
course-over-ground on boards without a magnetometer.

Against `deploy/apps/README.md`'s safety checklist: no TX, no network, no
identity or key access, no dynamic code loading; it persists only calibration
and unit choices, and writes only on a user action (the store flushes at close).
Bounded loops, no busy-waits; worst tick measured at ~11k of the 100k instruction
budget.

Deliberately **not** seeded into `lua_builtin.h`: that only applies to
`CAP_BUILTIN_LUA_APPS` boards and turning it on removes the Store's Apps tab
there.

## Testing

Verified on a ThinkNode M9 throughout: driver bring-up, calibration, the axis
measurement above, map timings, and app open/close (`leaked=0`, PSRAM returns to
within 44 bytes across a cycle). The app is also exercised by a host harness
built from the vendored Lua with the same `LUA_32BITS` numeric model, covering
six board geometries, both font scales, calibration recovery and the input model.

**After merging, `scripts/deploy-apps.sh` still needs running** for the app to
appear in the Store — merging the catalog entry does not publish it.

## Note

`sys.gps()`'s altitude is beta_68's `alt_m` only. The merge first carried `alt`
alongside it to avoid breaking a shipped app, but gpscompass has never been
published, so the duplicate key was dropped and the app reads `alt_m`.


cvhviz:
Ready for review whenever you have time, and rebased onto `main` as of 7ab0d69 (beta_68 plus today's tall-bar and Discord batches) — the SDK conflicts are resolved keeping both sides, so `ui.text_w` / `ui.text_lines` and the `measure` cap are all intact alongside the new sensor caps.

Verified on real M9 hardware throughout, and the compass now reads correctly — including the part that took the longest, which was tilt: at this latitude the field dips ~60°, so a two-axis heading moves ~1.5° for every 1° the device is tipped, and that alone was tens of degrees hand-held. Driving the QMI8658 and rotating the field back into the horizontal plane fixed it.

Both sensors' axis orientations were **measured**, not guessed, since nothing documents them for this board:

| | magnetometer (QMC6309) | accelerometer (QMI8658) |
|---|---|---|
| +X | left edge | right edge |
| +Y | top edge | top edge |
| +Z | into the screen | into the screen |

Worth flagging for anyone else touching these parts: the M9's hard-iron bias is about **7× Earth's field** (≈ −0.34 / −3.4 / −2.8 G against a 0.26 G horizontal signal), which is why an uncalibrated device barely moves the dial and why the driver runs at ±32 G. It also means calibration has to be done by rotating the device **in one place** — carrying it around while turning also translates it through the field of a desk or laptop, which corrupts the fit by more than half the signal. The app now measures coverage from gravity and refuses a sweep that never turned the device over.

Two things in here are not M9-specific and may be worth a look on their own:
- `luaNetWorkerService()` cleared `s_net_pending` before making its HTTP request, so an app closing during a round trip freed the buffers under the worker. The flag now spans the request.
- Leaving the Map tab freed all nine decoded tiles, so every re-open re-read and re-decoded nine JPEGs — measured at ~2.5 s on every visit, now 245 ms on boards with the PSRAM to keep them (the 2 MB V4 keeps the old behaviour).

One request for after merge: **GPS Compass will not appear in the Store until `scripts/deploy-apps.sh` runs** — the catalog entry is in the repo but publishing is the rsync step, which needs the VPS credentials.

B0N3-GD:
Merged to `main` as e3c9f9b. Thorough work, and the parts that were measured rather than assumed are the parts that make it trustworthy: the axis table at four headings, the hard-iron bias explaining the ±32 G range, and the before/after map timings.

Three changes on the way in, all small:

**Tile-keep gate polarity.** `kPsramTotalB && kPsramTotalB < 4 MB` puts a board reporting zero PSRAM on the roomy side of the test, which is backwards for the one case that can least afford it. Dropped the non-zero guard so zero frees like the V4 does. No board in the matrix reports zero today, so this is only about the next one.

**gpscompass seeding.** You deliberately kept it out of `lua_builtin.h`, and that was right: it is 55 KB of Lua, more than every other app combined, and it wants a magnetometer the seeded boards do not have. Baked in, it took the V4 from 89.2% to 90.6% flash. The problem was that the intent lived only in whether someone remembered not to regenerate the header. It is now `"seed": false` in `apps.json`, honoured by the generator and documented in `deploy/apps/README.md`, because that generator runs from a pre-build hook as of this branch and would have quietly baked your app in on the next build.

**One link break, mine not yours.** `consoleModeToggleCb` was defined inside a `!HAS_TANMATSU` region while the Settings row that binds it compiles everywhere, so your branch was the first thing to build the Tanmatsu since and took the blame. Moved out; the console boot path is gated on `CAP_CONSOLE` alone, so the toggle now works there rather than just appearing.

Built on all seven S3 envs plus both ESP32-P4 targets. `deploy-apps.sh` still to run for the app to appear in the Store, as you noted.

PR #319, opened 2026-08-24
Feature: "At a glance" idle-dim notification overlay

codemonkeybr:
## Summary

New, board-agnostic notification style, distinct from the T-Deck's existing
opt-in "msgFlash" (which fully wakes to the live app + pulses the keyboard
backlight): while the device is unlocked but its screen has idle-dimmed to
off, an incoming message (DND permitting) briefly lights a plain black
overlay showing the sender/channel name and the message text, then auto-dims
after a fixed window on its own. It never opens the chat, so the message
stays unread and whatever tab/chat was open underneath is left exactly as it
was.

## Where the toggles live and how they work

Both toggles are in **Options > Display**, stacked together (the
while-locked switch used to live under Options > Lock screen; this PR moves
it to sit with the feature it modifies):

1. **"At a glance"** — master switch for the whole feature.
   - Backed by `touchPrefsGetGlanceEnabled()` / `SetGlanceEnabled()`
     (`TouchPrefsStore`, NVS key `glance_en`), **default ON** — matches the
     previously-unconditional behavior before this toggle existed.
   - Turning it off skips the overlay entirely in
     `UITask::newMsgImpl()`; turning it back on restores whatever the
     while-locked switch was previously set to (see below) — it doesn't
     reset that preference.
   - While the master is off, the while-locked switch below is visually
     disabled (`LV_STATE_DISABLED`) so it can't be edited, but its stored
     value is untouched.

2. **"At a glance while locked"** — secondary switch, gates whether the
   overlay also fires while the device is manually/idle **locked**, not
   just unlocked-and-idle-dimmed.
   - Backed by `touchPrefsGetGlanceWhenLocked()` / `SetGlanceWhenLocked()`
     (NVS key `glance_lck`), **default OFF** — this is a real trade-off
     (message text becomes readable off a locked device), not just a
     convenience default, so it stays opt-in even when the master is on.
   - Only editable while the master switch is on.

Net effect: with the master on and while-locked off (the default), the
overlay fires only when the device is unlocked but idle-dimmed. Enabling
while-locked additionally shows it on a manually/idle-locked device.
Disabling the master suppresses the overlay outright regardless of the
while-locked setting.

## Test plan

- [x] Build `heltec_v4_tft_companion_radio_usb_tcp_touch` and
      `LilyGo_TDeck_companion_radio_touch`, confirm both compile clean.
- [x] Build the Tanmatsu (IDF-native, `tanmatsu/` app, not a PlatformIO env)
      and verify on hardware — same shared `UITask.cpp`/`TouchPrefsStore`
      code path, no board-specific divergence for this feature.
- [x] On device: unlock, let screen idle-dim, send a message from another
      node — overlay shows, auto-dims after its window, chat stays unread.
- [x] Toggle master off — send another message while idle-dimmed — no
      overlay.
- [x] Toggle master back on — confirm while-locked switch's prior state
      (not reset) and its editability track the master switch.
- [x] With master on, enable while-locked, manually lock the device, send a
      message — overlay shows on the locked screen.
- [x] With master on, while-locked off, manually lock, send a message — no
      overlay.


D-Melhede:
The summary described the function as "overlay showing the sender/channel name and the message text, then auto-dims".
The actual implementation shows channel name + message, but not the sender.
I would like to be able to see who said what was in the displayed message, so you can also see if one of your friends is active and say hey.
Please add the sender name in addition to the channel name it was received in.
Also, either a configurable time to display the message or simply a longer time than now. I can't read long messages before the screen goes black again.

PR #320, opened 2026-08-24
Latch keyboard symbol modifiers

### LINKED ISSUE #300: [feature] Make the symbol and alt keys latch instead of requiring a hold
Requested on Discord by istvan0858, mambo_430 and jjsadv (#feature-requests, 11 to 15 Aug 2026).

The symbol key has to be held down while typing the symbol. The ask is to make it latch: press once, next keypress is a symbol, then it releases. Same for the alt key.

> at least the alt key could have this function. the alt key is underused anyway, or the same button works so you don't have to hold it down for a long time.

Suggested alternative binding, so a latch does not remove the existing hold behaviour:

> Or latch it by pressing the up-arrow and Sym.

Third reporter puts it more strongly, and it is worth quoting because it is about sustained use rather than a nicety:

> Is there a tech reason the keyboard isnt latching? Besides the keyboard just being a bad design, its the next major issue. its just a pain to use if you want to get close to typing good sentences.

Both behaviours can coexist: hold still works, a tap latches for one keypress, a double tap could lock until tapped again (the usual shift/caps pattern). Applies to the T-Deck and any other board with a hardware symbol key.

oumike:
## Summary

- Make a modifier tap select the symbol layer for the next key.
- Make a double tap lock the symbol layer until the modifier is tapped again.
- Preserve the existing held-modifier behavior and physical modifier chords.
- Apply the behavior to T-Deck and both T-LoRa Pager variants.

### T-Deck

- Hold **Sym** to type symbols as before.
- Tap **Sym** for one symbol; double-tap it to lock symbols; tap again to unlock.
- A solo **Alt** tap provides the same one-shot/locked symbol behavior.
- Held Alt keeps the existing base layer and preserves the controller's Alt+B and Alt+C shortcuts.

Current LilyGO keyboard-controller firmware exposes a five-byte raw matrix mode. The driver probes that mode at startup and performs modifier translation in the main firmware when available. Older keyboard-controller firmware automatically returns to its existing one-byte ASCII mode, so normal typing is not broken; Serial reports that the C3 firmware must be updated before latching is available.

### T-LoRa Pager

- Hold **Fn/Alt** to type symbols as before.
- Tap Fn/Alt for one symbol; double-tap it to lock symbols; tap again to unlock.
- Physical Fn+Shift, Fn+Backspace, and Fn+encoder gestures retain their existing behavior.

The former solo-Fn navigation shortcut conflicted with the requested latch and has been removed. The encoder remains the navigation control, while the visible main-screen mnemonic keys remain available.

## Implementation

A shared `LatchedModifier` state machine implements held, one-shot, locked, double-tap, and discarded-input transitions. Board-specific, hardware-independent state classes contain the exact T-Deck and Pager matrix maps and translation behavior, allowing the production event paths to be tested directly on the host.

T-Deck raw-mode detection uses LilyGO commands `0x03` and `0x04`. Raw frame translation, input-mode transitions, queue clearing, and cross-core publication are ordered under one critical section. Transition frames establish a physical-key baseline without publishing stale input. Legacy mode drains the controller's single latest-byte mailbox once across a UI-mode transition.

Modifier state is discarded while the screen is dark or manually locked and while Console or Remote Mode intentionally consumes input. A held modifier released after wake therefore cannot arm an unexpected latch. T-Deck and Pager key rings use critical-section-protected producer/consumer publication.

M9 remains unchanged because its keyboard controller emits only resolved characters and does not expose modifier transitions. Tanmatsu Alt remains its existing accent-composition workflow rather than being redefined as a symbol layer.

## Testing

Host state-machine tests passed under both C++11 and C++17 with `-Wall -Wextra -Werror`:

- shared one-shot, lock/unlock, discard, and timer-wrap behavior
- T-Deck held/tapped/locked Sym and Alt, Shift, Alt+B/Alt+C, empty mappings, and discard/release behavior
- Pager held/tapped/locked Fn, Shift/Caps, Fn+Shift, Fn+Backspace, Fn+encoder equivalent, Space/Backspace holds, and discard/release behavior

Firmware builds passed:

```text
LilyGo_TDeck_companion_radio_touch             SUCCESS
heltec_v4_tft_companion_radio_usb_tcp_touch    SUCCESS
tlora_pager_lr1121_companion_radio_touch       SUCCESS
tlora_pager_sx1262_companion_radio_touch       SUCCESS
```

T-Deck build size:

```text
RAM:   32.0% (104900 / 327680 bytes)
Flash: 78.5% (3190013 / 4063232 bytes)
```

The T-Deck upload task also completed successfully. `git diff --check` passes. Existing LVGL and RadioLib warnings remain unchanged.

## Hardware verification

Before merge, verify on a raw-capable T-Deck and both Pager radio variants:

1. Hold modifier plus multiple keys.
2. Tap modifier, type one symbol, then type a base-layer key.
3. Double-tap modifier, type several symbols, then tap to unlock.
4. Exercise the existing held Alt/Fn shortcuts.
5. Repeat around screen-off, lock/unlock, and Remote Mode transitions.
6. Confirm legacy T-Deck controller firmware logs fallback mode and retains normal ASCII typing.

Closes #300

### DIFF
--- a/DEVICES.md
+++ b/DEVICES.md
@@ -23,6 +23,10 @@ USB) and the [GitHub releases](https://github.com/ALLFATHER-BV/wadamesh/releases
 - **T-Deck**: the everything device: touch, physical keyboard, trackball cursor
   or d-pad navigation, microSD (deep 5000-message chat history, map tile packs,
   data storage), GPS on the Plus, notification sounds through the I2S speaker.
+  Tap Sym or Alt for one symbol, or double-tap either to lock the symbol layer;
+  this needs [LilyGO keyboard-controller firmware with raw matrix mode](https://github.com/Xinyuan-LilyGO/T-Deck/tree/master/examples/Keyboard_ESP32C3)
+  (June 2025 or newer). Older controller firmware keeps normal typing and
+  reports the unavailable latch mode on Serial.
 - **Heltec V4 + TFT**: touch UI with the on-screen keyboard; the optional
   Expansion Kit adds environment sensors (home-screen chart) and a piezo
   buzzer. V4.3 boards get the switchable high-gain receive LNA toggle.
@@ -33,6 +37,8 @@ USB) and the [GitHub releases](https://github.com/ALLFATHER-BV/wadamesh/releases
 - **T-Lora Pager**: no touchscreen at all — the QWERTY keyboard and the rotary
   encoder drive everything (Alt+turn free-scrolls a page; see
   [TLORA_PAGER_SHORTCUTS.md](TLORA_PAGER_SHORTCUTS.md) for the full key map).
+  Tap Fn/Alt for one symbol or double-tap it to lock the symbol layer; physical
+  hold combinations keep their existing behavior.
   GPS, microSD, keyboard backlight, lock screen and notification sound through
   the onboard codec and amp all work. Two builds, one per radio: LR1121 and
   SX1262 — flashing the wrong one leaves you with no radio, so check the label
--- a/TLORA_PAGER_PORT.md
+++ b/TLORA_PAGER_PORT.md
@@ -225,8 +225,8 @@ WASD-panning).
    `updatePagerAltShiftChord()` toggles Caps Lock ONLY while a text field is
    actually being edited (same "ta" derivation `handleHwKey()`'s TLORA_PAGER
    branch already uses to tell a bound-but-unfocused field apart from one
-   actually being edited), and jumps straight to Home (`navGoToMainTab`)
-   everywhere else.
+   actually being edited), and is a no-op everywhere else. Alt+Backspace is
+   the context-independent Home shortcut.
 3. **@-mention contact picker made encoder-navigable** (commit `57802eb`) —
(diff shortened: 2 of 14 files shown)


jrote1:
I'm looking forward to testing this, I was just about to raise an issue for this feature

jrote1:
I've built and installed this PR and can't seem to get it to work, I also installed this firmware on the keyboard https://github.com/hreikin/tdeck-keyboard is this the correct one?
Thanks

Danie10:
OK it is not supposed to work on the M9 then? It would be great for this to work there too, but on shift and sym keys for the same reasons.

Danie10:
@oumike this is not working yet in beta 76 for the M9. Was this supposed to also work for the M9, or must I log a new issue for that?

oumike:
I’ll take a look. On Sep 4, 2026, at 08:23, Danie ***@***.***> wrote:﻿Danie10 left a comment (ALLFATHER-BV/wadamesh#320)
@oumike this is not working yet in beta 76 for the M9. Was this supposed to also work for the M9, or must I log a new issue for that?

PR #335, opened 2026-08-27
Correction for the Battery reading for the Lilygo T-Deck

alphalord336:
This fixes the issue where the battery was incorrectly reading only 4.09 V after a full charge.


B0N3-GD:
Merged. Reading 4.09 V on a fully charged pack is a real complaint and the direction is clearly right.

One question so we can get the constant exactly rather than approximately: **what does a multimeter read at the battery terminals** when the device reports that 4.09 V? `analogReadMilliVolts()` already applies the ESP32's own per-chip ADC calibration, so what is left is the board's resistor divider, which is the same on every T-Deck. That means a good measurement from one device fixes it for all of them, and a guess from one device is wrong on all of them.

Your 2.037 puts a 4.09 V reading at 4.166 V. If the pack is actually sitting at 4.20 V just off the charger, the true ratio is nearer 2.054. Both are better than 2.0, so this is an improvement either way and I would rather ship it than leave everyone reading low.

If you have a meter to hand, post the pair (reported vs measured) and I will set it precisely.

PR #346, opened 2026-08-28
attaky: wire up the front D-pad for UI navigation

oumike:
The Attaky Mesh Deck has nine onboard buttons, but only POWER was wired up — the
front D-pad and SELECT did nothing. This binds them to the UI.

## Button mapping

Taken from Attaky's hardware catalog: the front buttons hang off `AW9523 @0x59`
(P00–P07, active-low), not direct GPIO.

| Pin | Button | Action |
|---|---|---|
| `P04` / `P00` | UP / DOWN | Move focus; page-scroll when focus has nowhere left to go |
| `P01` / `P03` | LEFT / RIGHT | Move focus; switch tabs on the tab bar; walk the caret in a field being edited |
| `P02` | SELECT | Activate the focused item; the first press on a focused field starts typing |
| `P05` / `P06` | L1 / R2 | Read, but deliberately left unbound — nothing claims them yet |
| `P07` | POWER | Unchanged (panel toggle; the long-press power-cut stays below firmware) |

## Driver

`AttakyMeshSeriesKeys` went from testing a single bit to debouncing the whole P0
byte and queueing press edges onto a small ring.

- **Auto-repeat** on the four directions after 450 ms at ~130 ms, so a long
  contact list doesn't cost one press per row. SELECT never repeats — one press
  must stay one activation.
- **All-bits-low glitch guard.** An `0x00` read means every button is down at
  once, which is a bad read, not a hand. Previously that would have fired a
  spurious panel toggle; now it resyncs the debouncer instead of treating the
  recovery as presses.
- Failed I²C reads still hold the previous state rather than being mistaken for
  `0xFF` ("nothing pressed"), as before.

## UI wiring

`CAP_KEYPAD_NAV` is now on for the board, so the D-pad drives the same focus
group the Tanmatsu arrows and the T-Deck trackball already drive — no per-screen
handling needed. `attakyNavPump()` is this board's pump, draining the expander
queue in `UITask::loop`.

It carries the captures the other keypad boards need — an open dropdown owns
UP/DOWN (moving group focus would close it), and the @-mention picker owns them
too since its cells are `NAV_SKIP_FLAG` and unreachable by focus. It is
deliberately narrower than the M9's: that board has no touch, so it also needs
d-pad paths for Back, Home, the chat-bubble action menu and map panning, all of
which are still a tap away here.

With the screen dark a press only wakes and is swallowed, so waking can never
also activate whatever the focus ring happened to be sitting on.

## Two notes for reviewers

- **The Attaky is the first `CAP_KEYPAD_NAV` board with `CAP_KEYBOARD == 0`,**
  which broke an assumption. `hwKeyDismissTopPopup()` was gated on
  `CAP_KEYBOARD`, but `navGoToMainTab()` inside the keypad-nav block calls it —
  so the old gate left the declaration undefined. Widened to
  `CAP_KEYBOARD || CAP_KEYPAD_NAV`, with a note left in `device_caps.h` so the
  next board here doesn't hit the same edge.
- **Nav is forced on at boot rather than read from `touchPrefsGetKbdNav()`.** The
  "Keyboard navigation" settings switch is `CAP_TRACKBALL`-only, so the pref has
  no UI on this board; reading it would let a stale value leave soldered-on
  buttons dead with no way to turn them back on.

## Testing

Built clean on `attaky_mesh_series_companion_radio_touch`. Also built
`ThinkNode_M9_companion_radio_touch` as a regression check, since the
`hwKeyDismissTopPopup` gate and `CAP_KEYPAD_NAV` are shared. Both pass on the
tree merged up to current `main` (beta_72).

Not yet exercised on physical hardware — the debounce and auto-repeat timings are
the part most worth a real-device sanity check.


B0N3-GD:
Merged. Nine buttons on the board and one of them wired up was an odd state to leave it in.

The part I checked hardest is the one you flagged yourself: this makes the Attaky the first board carrying `CAP_KEYPAD_NAV` with no physical keyboard, so anything behind that flag assuming a keyboard would break it. Built clean on all eight S3 environments and both ESP32-P4 targets, so nothing does.

Reading the expander queue in its own pump rather than pretending the buttons are GPIO is the right shape, and matches how the other nav boards differ only in what feeds the generic machinery.

PR #384, opened 2026-09-02
Fix direct-message timestamps

### LINKED ISSUE #374: Incorrect timestamp in sent message packet
Bug observed on a T-Deck Plus since at least beta_48.  Still happening on latest install of beta_74

The 'sent' timestamp in the message packet gets set incorrectly.  

- The system time given by the `clock` command and displayed on the screen is correct.  
- Using the `Sync clock time from system` menu option does not fix the problem.
- Rebooting/power cycling  multiple times will eventually fix the problem.  The reboot after an update usually brings the problem back.
- I suspect this also causes login to my room server to fail.

Image

oumike:
## Summary

Fixes outbound direct-message packets being stamped with arbitrary future dates after boot. The issue was caused by MeshCore's direct-message uniqueness state treating a random 32-bit value as a Unix timestamp floor.

Closes #374.

## Root cause

`core-v1.17.4` initializes `BaseChatMesh::_txt_last_ts` from unrestricted RNG state at boot. `sendMessage()` then obtains the correct RTC time but raises it above that random value before serializing the timestamp into the encrypted TXT payload.

The packet attached to #374 contains timestamp `3135105232`, or 2069-05-06 UTC, while the device clock is correct. At the current epoch, a random 32-bit seed exceeds real time on roughly 58% of boots, matching the reported behavior where repeated restarts eventually produce a good boot and an OTA restart can bring the fault back.

The normal clock display and `Sync clock from system` cannot repair this because they update the real RTC, not the private random `_txt_last_ts` watermark.

Room posts use the same direct plaintext path. A future-dated post can advance a room server's per-client replay watermark, causing later Join, request, or keep-alive traffic to be rejected as stale.

## Changes

- Adds an idempotent, fail-closed build patch for pinned MeshCore `core-v1.17.4`.
- Removes only the unrestricted random `_txt_last_ts` seed; other RNG-derived diagnostic state remains unchanged.
- Leaves `getCurrentTimeUnique()` and the within-boot `_txt_last_ts` ratchet intact, so direct messages remain monotonic while carrying real Unix time.
- Verifies the complete RTC-to-ratchet-to-serializer source contract before applying or accepting the patch.
- Rejects missing, duplicated, partially patched, marker-only, or version-drifted core sources.
- Handles existing PlatformIO dependencies before compilation.
- On a fresh dependency fetch, patches at pre-link and deliberately fails that first link so an unpatched object cannot ship; the next build compiles the corrected source.
- Registers the patch for all PlatformIO touch target families, including the inherited Heltec V4-R8 environment.
- Applies and verifies the patch when vendoring MeshCore for Tanmatsu.
- Requires T-Display P4 to use the canonical Tanmatsu-shared core, replacing stale standalone generated trees during dependency fetch and failing CMake configuration for a wrong link or unverified source.
- Persists the live `ClockFloorRTC` floor before controlled non-factory reboots, download mode, backup-import restart, SD migration restart, and deep sleep.
- Leaves factory reset excluded because it intentionally erases identity and replay state.

## Compatibility

The encrypted TXT wire layout is unchanged. Timestamp, attempt bits, text, ACK derivation, retries, and packet framing retain their existing format. Core-managed retries remain exact packet clones.

This is carried as a temporary dependency patch because both the pinned tag and current `ALLFATHER-BV/meshcomod` main still contain the defect and no corrected core tag exists. It can be removed when the same fix is released in a new core tag.

Already-poisoned room-server client records may need to be deleted/reset once so their replay watermark returns to current time.

## Validation

All repository hardware targets compiled at commit `0b7afc2`.

PlatformIO (`scripts/build-all-targets.sh --pio`):

- `heltec_v4_tft_companion_radio_usb_tcp_touch`
- `heltec_v4_r8_tft_companion_radio_usb_tcp_touch`
- `attaky_mesh_series_companion_radio_touch`
- `LilyGo_TDeck_companion_radio_touch`
- `tlora_pager_lr1121_companion_radio_touch`
- `tlora_pager_sx1262_companion_radio_touch`
- `ThinkNode_M9_companion_radio_touch`
- `rak_tap_v2_companion_radio_touch`

ESP-IDF 5.5.1:

- T-Display P4 linked successfully; `application.bin` is 3,478,816 bytes with 17% of the app partition free.
- Tanmatsu compiled and linked successfully from a no-space worktree; `application.bin` is 3,160,752 bytes. Its final generic OTA `app_check_size` reports the expected 2 MiB partition overflow because Tanmatsu ships this binary as an AppFS application; the repository's app-store release script explicitly treats this check as non-fatal after rejecting real compiler/linker errors.

Additional checks:

- Patcher self-test covers exact patching, idempotency, source drift, duplicate blocks, poisoned markers, and strict verification.
- Exact pinned `BaseChatMesh.cpp` rejects verification before patching and passes afterward.
- Simulated existing and fresh PlatformIO/SCons dependency lifecycles pass.
- Tanmatsu/P4 vendoring and CMake success/failure guards pass.
- PlatformIO configuration parsing passes with all target hooks present exactly once.
- Python, C++, and CMake editor diagnostics are clean.
- Shell syntax checks and `git diff --check` pass.

Hardware packet capture was not performed. Abrupt power removal can still lose timestamp-floor progress made since the last durable preference snapshot; controlled reset paths now persist it explicitly.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -15,7 +15,7 @@ framework = arduino
 board = heltec_v4
 monitor_speed = 115200
 board_build.partitions = partitions_tft_touch.csv
-extra_scripts = pre:scripts/build/pre_gen_baked.py, pre:scripts/inject_wifi_env.py, merge-bin.py
+extra_scripts = pre:scripts/build/pre_gen_baked.py, pre:scripts/inject_wifi_env.py, pre:scripts/build/patch_meshcore_txt_timestamp.py, merge-bin.py
 
 build_flags =
   ; 1.17 core: ESP32Board.cpp includes <target.h> -> MicroNMEALocationProvider.h (header-only
@@ -330,7 +330,7 @@ framework = arduino
 board = attaky_mesh_series
 monitor_speed = 115200
 board_build.partitions = partitions_tft_touch.csv
-extra_scripts = pre:scripts/build/pre_gen_baked.py, pre:scripts/inject_wifi_env.py, merge-bin.py
+extra_scripts = pre:scripts/build/pre_gen_baked.py, pre:scripts/inject_wifi_env.py, pre:scripts/build/patch_meshcore_txt_timestamp.py, merge-bin.py
 
 build_flags =
   ; 1.17 core: ESP32Board.cpp includes <target.h> -> MicroNMEALocationProvider.h (header-only
@@ -516,7 +516,7 @@ framework = arduino
 board = t-deck
 monitor_speed = 115200
 board_build.partitions = variants/lilygo_tdeck/partitions_tdeck_touch.csv
-extra_scripts = pre:scripts/build/pre_gen_baked.py, pre:scripts/inject_wifi_env.py, merge-bin.py
+extra_scripts = pre:scripts/build/pre_gen_baked.py, pre:scripts/inject_wifi_env.py, pre:scripts/build/patch_meshcore_txt_timestamp.py, merge-bin.py
 
 build_flags =
   ; 1.17 core: ESP32Board.cpp includes <target.h> -> MicroNMEALocationProvider.h (header-only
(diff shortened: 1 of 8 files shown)


B0N3-GD:
Merged, closes #374. Seeding a semantic Unix timestamp from an unrestricted random uint32 and then ratcheting the real clock above it is a proper root cause, and the fail-closed idempotent patch is the right shape for something that has to survive a dependency fetch.

One note: the IDF builds verify rather than patch, and the vendored copy in my checkout predated the script, so both P4 targets stopped configuring with your `FATAL_ERROR`. That is the design working, and your message said exactly what to do. Running the patch step against the vendored trees cleared it.

Worth calling out because anyone with an existing checkout will hit it once, and the fix is `fetch-deps.sh` as the error says.

PR #385, opened 2026-09-02
M9 battery pass: reachable battery saver, GPS-off leakage, mid-session GPS UART ring (+ spectrum sweep wall-clock bound)

cvhviz:
Two commits on a clean beta_75 base.

## M9 battery pass (`d59ce2c`)

A battery-focused audit of the M9 with three fixes:

1. **The battery saver was unreachable on the M9.** The touchSleep machinery (hooks, gates, pref restore) was fully wired under `HAS_TOUCH_UI`, but the only enable UI — the Settings→Battery switch — was compiled for the T-Deck and V4-R8 only, and the pref defaults OFF with no other setter. Both `#if`s now include `HAS_THINKNODE_M9`. The M9 is a natural fit: the throttle is a plain vTaskDelay in the loop task, wake is the keyboard poll that runs through it (≤50 ms latency on a dark screen), and the M9 sits in `batteryIsCharging()`'s `#else` branch (compile-time false), so the "USB powered" gate can never wrongly block. Matching the R8 precedent: no status-bar indicator; the settings row shows the live block reason.

2. **GPS-off burned Q16 base current on the M9.** `WadaNmeaLocationProvider` parked GPS_RESET *asserted* while stopped (the core's convention — free on a direct-wired active-LOW reset). The M9's reset is inverted (GPS_RST1 → R46 → NPN Q16, GPIO HIGH = asserted), so "asserted" sourced (3.3 V − Vbe)/R46 continuously whenever the GPS was OFF — its default state — for a module whose EN rail was already cut. The line now parks LOW: the zero-current level for both circuit styles (LOW *is* asserted on direct-wired boards, so nothing changes there; the provider is M9-only today anyway). Every start path still calls `begin()` then `reset()`, so the clean-reset-at-power-on guarantee holds.

3. **Mid-session GPS enable ran the whole session on the stock 256 B Serial1 ring.** The 4 KB TTFF ring is boot-gated on the persisted `gps_enabled`, and `setRxBufferSize` is a no-op on a running UART — so the session that first enables GPS had ~22 ms of RX slack at the M9's 115200 baud, and with the saver newly reachable, a single 50 ms park clips a whole NMEA burst (host-side: checksum-failed sentences discarded, stale fix until reboot). New `gpsEnsureBigRxRing()` in main.cpp cycles the idle-enough UART (end → 4 KB → begin at the initBasicGPS pins/baud) from BOTH "gps on" funnels — `UITask::toggleGPS()` and the companion `CMD_SET_CUSTOM_VAR("gps")` handler — right after a *successful* start, so a refused enable (GPS-less V4/R8 detect path) never spends the DRAM, and the enable test matches `setSettingValue`'s real semantics (anything but "0").

Also corrects both stale "keyboard S2 on the switched rail" comments (schematic truth per platformio.ini: always-on 3V3 via R3 — it decides the deep-sleep drain budget).

Full record, rationale, and the on-device VERIFY list in `variants/thinknode_m9/M9_PORT.md` → "Battery pass (2026-09-02)". Note for testers: the saver's proof is the Battery chart's *Idle power-save counters moving* (the CPU-at-80 readout is a decoy — that drop belongs to the pre-existing screen timeout), and the gate needs BLE + Wi-Fi actually off.

## Spectrum sweep wall-clock bound (`eb17bcc`)

Carried forward from the audit-pass-3 branch (everything else from that branch is already in beta_75): bounds the spectrum sweep tick by wall clock and records the real per-bin cost.

## Verification

- All 8 PlatformIO envs compile on the beta_75 base.
- The battery pass went through two adversarial multi-agent review rounds; every confirmed finding (including the UART-ring interaction, found by the review of fix 1) is addressed in this diff.
- Flashed and running on an M9 (hash-verified upload); on-device VERIFY items listed in M9_PORT.md.

### DIFF
--- a/src/MyMesh.cpp
+++ b/src/MyMesh.cpp
@@ -5003,6 +5003,21 @@ void MyMesh::handleCmdFrame(size_t len) {
         #if ENV_INCLUDE_GPS == 1
         // Update node preferences for GPS settings
         if (strcmp(sp, "gps") == 0) {
+          // Companion-app "gps on" mid-session: same 4 KB RX-ring upgrade the
+          // touch UI's toggleGPS does — the boot-time gate in main.cpp only
+          // installs it when gps_enabled was already persisted, and
+          // setRxBufferSize is a no-op on the running UART. Matches
+          // setSettingValue's own enable test (anything but "0" starts the
+          // GPS — not just "1"), and runs only on SUCCESS so a refused enable
+          // (no GPS detected on the V4/R8 detect path) doesn't pay the 4 KB.
+          // Cycling just after start is benign: everything here runs on
+          // loopTask, and the module has barely powered up (EN-pin boards) or
+          // at worst loses one torn, checksum-rejected sentence (EN-less
+          // boards, whose modules stream even while "stopped").
+          if (strcmp(np, "0") != 0) {
+            extern void gpsEnsureBigRxRing();
+            gpsEnsureBigRxRing();
+          }
           _prefs.gps_enabled = (np[0] == '1') ? 1 : 0;
           savePrefs();
         } else if (strcmp(sp, "gps_interval") == 0) {
--- a/src/helpers/WadaNmeaLocationProvider.h
+++ b/src/helpers/WadaNmeaLocationProvider.h
@@ -16,7 +16,9 @@
 // wada.sys.gps() binding then reports speed_kmh/course on that board.
 //
 // Kept in lock-step with the core header it mirrors (meshcomod core-v1.17.4);
-// if the core provider gains behaviour, port it here too.
+// if the core provider gains behaviour, port it here too. ONE deliberate
+// deviation (2026-09-02 battery pass): the reset line is parked LOW — not
+// asserted — while the GPS is stopped; see the comment in the constructor.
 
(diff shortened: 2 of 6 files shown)


B0N3-GD:
Merged. The battery saver being fully wired but with its only enable switch compiled for two other boards is a good catch, and exactly the kind of thing that reads as "feature missing" rather than "feature unreachable".

One thing to know for next time, and it cost both ESP32-P4 targets their link:

`gpsEnsureBigRxRing()` is called from `UITask.cpp` and `MyMesh.cpp`, which every target compiles, but it is **defined in `src/main.cpp`**, which the Tanmatsu and T-Display P4 never compile. They are ESP-IDF builds with their own `main.cpp`. So the S3 matrix stays green and both P4 boards fail at link with `undefined reference to gpsEnsureBigRxRing()`.

Fixed on merge with a no-op definition for those two boards in `UITask.cpp`. A no-op rather than a stub, because neither P4 drives its GPS through Arduino's `Serial1`, so there is no stock 256 byte ring to enlarge.

The standing rule, since this has bitten more than once: anything shared between the UI and the Arduino main belongs in `UITask.cpp`, and `src/main.cpp` may only `extern` it. Worth building one P4 target before a PR that adds a cross-file symbol:

```
cd tdisplay_p4 && ./build.sh build
```

PR #391, opened 2026-09-03
Add Seeed Wio Tracker L2 support

oumike:
Adds the Seeed Wio Tracker L2 as a build target. Seeed starts selling the board next week.

## What's here

- **New board**: `boards/wio_tracker_l2.json` — ESP32-S3, 16 MB flash, 8 MB octal PSRAM.
- **New variant**: `variants/wio_tracker_l2/` — `WioTrackerL2Board` (battery ADC through the expander's sense rail, SD presence), `WioTrackerL2Display` (quad-SPI panel via LovyanGFX, fixed 320x240 landscape, backlight through the expander), `WioTrackerL2Io` (I/O expander: LCD / GNSS / SD / audio-PA power rails, wake button, battery sense), `WioTrackerL2Touch` (GT911).
- **New env**: `[env:wio_tracker_l2_companion_radio_touch]` — SX1262 at 22 dBm, LVGL touch UI, GPS, microSD via SD_MMC, multi-transport companion.

## Board bring-up notes

- **GPS wiring**: the env hit the same `Serial1.setPins()` naming trap the V4-R8 and Pager envs carry — Arduino's `setPins()` takes `(rx, tx)`, so `PIN_GPS_TX` is the pin we *receive* on. The pins were copied from another env during bring-up and left the UART listening on GPIO17 while the module talks on GPIO18, so GPS never produced a fix. Confirmed on hardware: NMEA on GPIO18 at 9600 with 9 sats in view, GPIO17 silent at 9600/38400/115200. The comment in `platformio.ini` documents the trap.
- **Storage**: the microSD is SD_MMC (not Arduino `SD`), mounted at boot at 20 MHz and adopted by DataStore, so identity, prefs, contacts and chat history live on the card. The file manager's lazy remount matches that clock rather than re-`begin()`ing at Arduino's 40 MHz default, and it will not `end()` or power down a card the store adopted — one transient read error would otherwise silently kill every later write for the rest of the boot.
- **Panel**: fixed 320x240 landscape (`CAP_ROTATABLE 0`), so the keyboard's rotate-for-landscape arrows are not built on this board — they would turn a landscape keyboard into a portrait one.
- **Buttons**: either physical button wakes; a two-second hold turns the display off without hard-locking touch. The wake button hangs off the I/O expander, so it is polled alongside `PIN_USER_BTN`.
- **LovyanGFX**: `scripts/build/patch_lgfx_dmadesc.py` guards the ESP32 SPI DMA descriptor allocation against OOM — LovyanGFX frees the working descriptor array before allocating its replacement, and walks a null array when that allocation fails. The patch is per-env (each PlatformIO env owns its own libdeps copy) and verified again at link time.

## Scope

The branch also carries the accumulated touch-UI work from this fork's `square` development line — roughly 215 of the 236 `UITask.cpp` hunks are that, not L2-specific. Happy to split the board support onto a clean branch off `main` if you would rather review it on its own.

## Verification

Builds clean: Flash 75.7% (3,076,081 / 4,063,232 B), RAM 1.1%. Hardware bring-up is done for display, touch, GPS, microSD, battery and buttons; the most recent flash attempt was interrupted by a USB drop, so the current binary is not yet confirmed running on-device.


B0N3-GD:
Merged, thank you. Seeed sell this next week and you have the hardware, so the board bring-up is taken on your word: display, touch, GPS, microSD, battery and buttons are yours, confirmed on-device.

What this side could actually verify is that it costs nobody else anything, so the whole matrix was built: all nine S3 envs including the new one, plus both ESP32-P4 targets. All green.

On the development-line work you flagged and offered to split out: kept as-is. The three-way merge came through with zero conflicts and I checked explicitly that nothing recent was reverted rather than assuming it, since the diff against main showed 1706 deletions purely because the branch predates the last few merges. The prefs schema is still v54 and the recent additions are all still present.

One thing the PR could not do by itself: the board built but would never have shipped. Three lists decide that and none of them knew about it, so I added it to `release.sh` ENVS, to `gen-flasher-meta.py` BOARDS and to the flasher page, as `wadamesh-wio-tracker-l2` in its own Seeed Studio section. Beta channel only for now, the way the Attaky started, since stable is still beta_65. Worth knowing for the next board.

The `Serial1.setPins()` note in your description is a good catch and it is the same trap the V4-R8 and Pager envs carry. It is in the merge commit message so it is findable next time somebody stares at a silent GPS pin.

You also mentioned the last flash was interrupted by a USB drop and the current binary is not confirmed running. It is in the next beta either way, but if you get a clean flash before then and anything looks off, say so and it can go in before the build rather than after.

oumike:
I was able to get it running no issue, my usb cable however took a dive. 🤓On Sep 3, 2026, at 17:59, Kaj ***@***.***> wrote:﻿B0N3-GD left a comment (ALLFATHER-BV/wadamesh#391)
Merged, thank you. Seeed sell this next week and you have the hardware, so the board bring-up is taken on your word: display, touch, GPS, microSD, battery and buttons are yours, confirmed on-device.
What this side could actually verify is that it costs nobody else anything, so the whole matrix was built: all nine S3 envs including the new one, plus both ESP32-P4 targets. All green.
On the development-line work you flagged and offered to split out: kept as-is. The three-way merge came through with zero conflicts and I checked explicitly that nothing recent was reverted rather than assuming it, since the diff against main showed 1706 deletions purely because the branch predates the last few merges. The prefs schema is still v54 and the recent additions are all still present.
One thing the PR could not do by itself: the board built but would never have shipped. Three lists decide that and none of them knew about it, so I added it to release.sh ENVS, to gen-flasher-meta.py BOARDS and to the flasher page, as wadamesh-wio-tracker-l2 in its own Seeed Studio section. Beta channel only for now, the way the Attaky started, since stable is still beta_65. Worth knowing for the next board.
The Serial1.setPins() note in your description is a good catch and it is the same trap the V4-R8 and Pager envs carry. It is in the merge commit message so it is findable next time somebody stares at a silent GPS pin.
You also mentioned the last flash was interrupted by a USB drop and the current binary is not confirmed running. It is in the next beta either way, but if you get a clean flash before then and anything looks off, say so and it can go in before the build rather than after.

Issue #389, opened 2026-09-03
M9 lockscreen font color not working


Buko84:
There is a bug that color of the lockscreen font goes immediately black on wake.. see the attached video

https://github.com/user-attachments/assets/4f6ba972-d750-4039-bcd9-0e24df2a65d1

B0N3-GD:
I have not been able to reproduce this from the code, and I do not have an M9, so I need one detail from you to place it.

What I ruled out, so the answer is worth your time:

- There are exactly two places that set the lock clock colour, when the screen is built and on each clock update, and both read the same preference. So they cannot disagree with each other.
- None of the eight colour swatches is black, so a mis-selected swatch cannot produce it.
- The default is a soft white, so an unset or freshly defaulted preference is not black either.
- There is no fade, no opacity animation and no recolouring anywhere on the lock screen, so nothing is dimming the text after it is drawn.

That leaves either the preference reading back as black on this board, or something repainting outside those two paths. These questions separate the two:

1. Is it only the clock, or do the unread line and the text under it go black as well?
2. Does it stay black, or come back on its own within a minute or so? The clock repaints on the minute, so "recovers at the next minute" and "stays black until you unlock" point at different halves.
3. Does it also happen if you leave the colour at the default and do not pick one at all?
4. Which colour did you pick when you saw it?

Worth knowing: the T-Lora Pager had what reads like the same complaint, and it was worked around by forcing white on that board rather than finding the cause. So this may well be one bug that has now been seen twice, and getting it properly this time is better than hardcoding a second board.

Buko84:
I will record it tomorrow so you can take a look. For me there is something wrong because going thru UI the highlights are white, and picking the color also, when you are picking the color the one that you are standing on is going white. I will do the video or pictures.

B0N3-GD:
Your second message changes this, and it turns out you have found a real bug that is not the one in the title.

"When you are picking the colour, the one you are standing on is going white" is exactly right, and it is now fixed in the next beta. Each swatch had its colour set only for its default state, which left the theme free to repaint it when focused. On a board driven by the d-pad the swatch you are standing on is always the focused one, so it turned white and hid the very thing you were choosing. The colour is now pinned across the focused, pressed and checked states, and focus is drawn as a ring outside the swatch rather than a fill over it, so a swatch always looks like its own colour.

That also means the picker has been effectively unusable by d-pad, which is worth keeping in mind when you retest: you may never have successfully set the colour you intended.

[...]

PR #396, opened 2026-09-04
Fix Send advert screen navigation and dropdowns, plus the DND icon

oumike:
Four small UI fixes, all reported on the ThinkNode M9 build. One file touched (`src/ui-touch/UITask.cpp`).

## Keypad nav could skip a row

Opening **Send advert** puts the cursor on **Flood**; pressing down skipped the **On boot** switch and landed on the **Flood every** dropdown two rows below.

`navMoveDir` ranked candidates by `primary + 8 * cross` — the gap along the pressed axis, plus a heavy penalty for being off-axis — with no notion of rows. The "Flood every" dropdown is wide enough to overlap the Flood button's x-range, so its cross penalty was 0, while the "On boot" switch directly beneath is narrow and right-aligned, so its horizontal gap out-weighed a whole extra row of distance. (Zero-hop happened to work: being on the right half, it overlapped the switch horizontally.)

`navMoveDir` now walks the candidates twice. The first pass finds the nearest element along the pressed axis and takes its rectangle as the row (or column) band; the second discards anything lying entirely past that band, then applies the existing score to what's left. A press can no longer jump over an intervening focusable, while alignment still decides *within* a row — candidates that merely overlap the band (a taller neighbour, a switch beside a button, a grid row) stay in the running. The rule is direction-generic, so it also stops LEFT/RIGHT skipping past a nearer column.

The shared geometry moved into two helpers, `navDirCandidate()` (valid / visible / not an `NAV_HMOVE_FLAG` gear) and `navDirMetrics()` (direction test plus the gap computation), so both passes use the same code. Existing comments carried over verbatim.

Down from Flood now goes **On boot → Flood every → Local every**, and the same in reverse.

## The focus highlight was invisible on the advert dropdowns

LVGL's stock theme in this build is the **light** one — nothing calls `lv_theme_default_init`, and `include/lv_conf.h` leaves `LV_THEME_DEFAULT_DARK` at its default — so a dropdown with no explicit styling renders as a near-white card. `navFocusCb` draws the keyboard cursor by reverse-videoing the focused widget to `COLOR_TEXT`, i.e. white, so on those two widgets it painted white on white and vanished.

The settings pages never showed this because they set `COLOR_PANEL` inline on every dropdown they build. The two on Send advert were created bare.

## An open dropdown list was clipped by the title bar

`lv_dropdown`'s constructor parents its list to `lv_obj_get_screen()`, which for a page living on `lv_layer_top` *is* that layer — so the list sat below the status bar in the child order (the advert page moves the bar to the foreground when it opens) and got covered where they overlapped. `clampDropdownListCb` already exists for exactly this case, with a comment saying to wire it onto every settings dropdown; these two were never wired.

## One helper for both dropdown fixes

`styleDropdown()` now does the whole standard treatment in one call:

- panel/text/border colours and the 12px font on the button and its open list;
- an accent-filled selected row with `COLOR_ON_ACCENT` text — LVGL draws the highlighted option in `LV_PART_SELECTED` with the `CHECKED` state, which the stock theme fills with its own blue primary rather than the UI accent;
- `clampDropdownListCb` wired to `LV_EVENT_CLICKED`.

Applied to the two advert dropdowns. The ~20 settings-page dropdowns keep their inline styling for now — they already look right; happy to sweep them onto the helper in a follow-up if the accent-coloured selected row is wanted everywhere.

## Do Not Disturb gets its usual crescent moon

The control-center DND chip used a bell that gained a slash when DND was on, which reads as "notification sound" and collided with the Sound chip right beside it. The status bar already used the moon.

The chip now renders `TOUCH_SYM_MOON` in both states — its own fill (solid accent on, faint off) already carries the state. No new font asset is needed: `sleepicons_font` carries U+F185/F186 and is already spliced into the `g_font_16` fallback chain that `ccToggle` renders its icon in. Two stale "bell" comments on `ccDndCb` updated to match; `TOUCH_SYM_BELL_SLASH` is now unused but its `#define` and glyph stay, since `TOUCH_SYM_BELL` is still referenced by the Lua icon-name map and the font is needed either way.

## Testing

Tested on hardware: T-Deck, ThinkNode M9, and T-LoRa Pager.

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -4000,6 +4000,34 @@ static const char* const kNavDirNames[8] = { "Up", "Down", "Left", "Right", "Sel
 // direction (not just next/prev in the linear list). Scores candidates in that
 // half-plane by primary-axis distance + a cross-axis penalty (prefers aligned + close).
 enum NavDir { NAV_UP, NAV_DOWN, NAV_LEFT, NAV_RIGHT };
+// Can `o` take focus from `cur` on a `dir` press at all? (Valid, visible, and not a
+// horizontal-only secondary target — chat-row gears are reachable by LEFT/RIGHT only, so
+// UP/DOWN walk the primary rows instead of hopping onto a gear.)
+static bool navDirCandidate(lv_obj_t* o, lv_obj_t* cur, int dir) {
+  if (!o || o == cur || !lv_obj_is_valid(o) || lv_obj_has_flag(o, LV_OBJ_FLAG_HIDDEN)) return false;
+  if ((dir == NAV_UP || dir == NAV_DOWN) && lv_obj_has_flag(o, NAV_HMOVE_FLAG)) return false;
+  return true;
+}
+// Geometry of candidate `b` relative to the focused rect `a`: true when it lies in `dir`,
+// with the gaps navMoveDir ranks by. Direction uses the CENTRE delta (which side the
+// candidate is on); the gaps are rect-to-rect, not centre-to-centre — `primary` is the gap
+// along the pressed axis (how far past the current edge it sits) so the element IMMEDIATELY
+// in that direction wins, and `cross` is the off-axis gap, 0 whenever the rects line up on
+// that axis (a full-width row below a right-aligned switch still counts as "straight down").
+static bool navDirMetrics(const lv_area_t& a, const lv_area_t& b, int dir, long* primary, long* cross) {
+  const int dx = (b.x1 + b.x2) / 2 - (a.x1 + a.x2) / 2;
+  const int dy = (b.y1 + b.y2) / 2 - (a.y1 + a.y2) / 2;
+  long hgap = b.x1 - a.x2; if (a.x1 - b.x2 > hgap) hgap = a.x1 - b.x2; if (hgap < 0) hgap = 0;
+  long vgap = b.y1 - a.y2; if (a.y1 - b.y2 > vgap) vgap = a.y1 - b.y2; if (vgap < 0) vgap = 0;
+  bool ok;
+  switch (dir) {
+    case NAV_UP:    ok = dy < 0; *primary = vgap; *cross = hgap; break;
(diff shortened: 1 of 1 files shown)


B0N3-GD:
Merged, thank you. Four real fixes and the write-up made them quick to check.

The navigation one is the valuable half. Scoring candidates on axis distance plus an off-axis penalty with no notion of rows was always going to let a wide element two rows away beat a narrow one directly beneath, and establishing the row band first is the right shape for it. That is shared code across every keypad board, so I built the whole matrix: all nine S3 envs including the new Wio Tracker L2, plus both ESP32-P4 targets. All green.

Your note about the stock light theme is worth keeping in mind. Reverse-videoing to white on a card the theme paints near-white is not really a bug in either half on its own, which is exactly why it only showed on the two widgets nobody had styled inline.

On the ~20 settings dropdowns you left on inline styling: yes to the follow-up sweep, and the accent-coloured selected row is wanted everywhere. Worth doing as its own PR so it can be looked at as a straight visual diff.

One thing that landed alongside your merge, since it is adjacent: Buko84 reported on #389 that the lock-screen colour swatches go white when you stand on them with the d-pad. Same shape of problem as yours, a widget whose colour was set only for the default state and then repainted by the theme when focused. Fixed in the same commit, with focus shown as a ring outside the swatch rather than a fill over it, so the swatch keeps being its own colour.

B0N3-GD:
Correction to my previous comment: when I posted it, this was not actually merged. It is now.

The merge had refused because my working tree still had an uncommitted change to the same file, and I had filtered the command output in a way that hid the error, so what got committed was only my own change under a message describing yours. GitHub kept showing the PR open, which is what made me look again.

Now genuinely merged, with both sides verified present in the tree, and the full matrix rebuilt on the real merge rather than on the commit that lacked it: all nine S3 envs including the Wio Tracker L2, plus both ESP32-P4 targets. All green.

Everything I said about the work stands. The row-band change is the valuable half and the write-up made it quick to check.

PR #413, opened 2026-09-05
Add authenticated browser file transfer

### LINKED ISSUE #400: File copying for ThinkNode M9
As from https://github.com/ALLFATHER-BV/wadamesh/releases#release-beta_70 we can now do screenshots but with the M9 not having the removal SD card like the T-Deck has, we need a way to copy files to/from the M9. Seems like a new app that does FTP or mounts a file share (macOS, Linux, and Windows) may be needed. An app may be perfect as it can be started or stopped when needed.

Or not sure if the Remote function could handle this in some way.

I thought otherwise it could mount the SD card when the USB data cable is plugged in, but that may interfere with the WADAMESH firmware that writes to it.

oumike:
## Summary

- add a six-digit authenticated browser file-transfer session with bounded WebSocket framing and cross-core backpressure
- upload safely through `/transfer/.upload.part` with exact offsets, CRC32 verification, and atomic commit
- list and download files from `/screenshots` and `/transfer`
- enable the app on removable-card targets using either Arduino `SD` or `SD_MMC`, while keeping no-card builds free of the tile and transfer payload
- integrate transfer activity with the existing SD lifecycle and app hide/show controls

## Supported storage targets

- Heltec V4-R8
- LilyGo T-Deck
- T-LoRa Pager LR1121 and SX1262
- ThinkNode M9
- Wio Tracker L2
- Tanmatsu
- T-Display P4

Heltec V4, Attaky Mesh Series, and RAK Tap V2 continue to compile without exposing File Transfer because they do not provide a suitable removable-card backend.

## Validation

- M9 hardware workflow tested successfully
- all 9 PlatformIO environments build successfully via `./scripts/build-all-targets.sh --pio`
- host protocol/capability test passes with `-Wall -Wextra -Werror`
- embedded browser JavaScript parses and matches the standard CRC32 test vector
- VS Code diagnostics and `git diff --check` are clean
- Tanmatsu and T-Display P4 storage/transport paths were statically reviewed; their ESP-IDF builds could not run because the project-local IDF toolchain is not installed on this machine

Fixes #400

### DIFF
--- a/src/helpers/esp32/MultiTransportCompanionInterface.cpp
+++ b/src/helpers/esp32/MultiTransportCompanionInterface.cpp
@@ -2,6 +2,10 @@
 #include <helpers/RepeaterTcpOtaEmit.h>
 #include "WifiRuntimeStore.h"   // persist BLE on/off (ble_en) across reboots
 #include "WebMirror.h"          // web UI mirror bridge (served over the WS server)
+#include "WebFileTransferConfig.h"
+#if WADA_WEB_FILE_TRANSFER
+#include "WebFileTransfer.h"
+#endif
 #include <esp_heap_caps.h>
 #include <string.h>
 
@@ -118,7 +122,12 @@ static void wsMirrorStreamTask(void* arg) {
   WebSocketCompanionServer* ws = static_cast<WebSocketCompanionServer*>(arg);
   for (;;) {
     ws->serviceMirror(g_web_mirror);
-    vTaskDelay(pdMS_TO_TICKS(g_web_mirror.clients() > 0 ? 2 : 25));  // fast when a browser is watching, idle otherwise
+#if WADA_WEB_FILE_TRANSFER
+    const bool browser_active = g_web_mirror.clients() > 0 || g_web_file_transfer.clients() > 0;
+#else
+    const bool browser_active = g_web_mirror.clients() > 0;
+#endif
+    vTaskDelay(pdMS_TO_TICKS(browser_active ? 2 : 25));  // fast when a browser is active, idle otherwise
   }
 }
 
--- a/src/helpers/esp32/WebFileTransfer.cpp
+++ b/src/helpers/esp32/WebFileTransfer.cpp
@@ -1,0 +1,203 @@
+#include "WebFileTransfer.h"
+
+#if WADA_WEB_FILE_TRANSFER
+
+#include <Arduino.h>
+#include <esp_heap_caps.h>
+#include <stdlib.h>
+#include <string.h>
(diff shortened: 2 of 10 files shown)


B0N3-GD:
Merged, and thank you for the care in this one. Reviewing an authenticated write path is different from reviewing a UI change, so here is what I actually checked rather than a rubber stamp.

The two things that would have made this unmergeable are both handled properly.

Path handling: `fileNameValid()` is an allowlist rather than a denylist, which is the right way round. `/` is not in the permitted set, so traversal cannot be expressed at all; a leading dot is rejected, which takes out `..` and hidden files together; the `.part` suffix is rejected, so the upload temp cannot be targeted; and `readablePath()` requires one of the two prefixes, validates the leaf, and then checks for `/` again anyway. Belt and braces on the part that deserves it.

Authentication: three failures then a thirty second lockout puts a brute force of a six digit code far out of reach. The comparison is not constant time, but a timing attack on an integer compare over Wi-Fi is not a practical concern, so that is a fair trade rather than an oversight.

I ran your host test with `-Wall -Wextra -Werror` and it passes.

The part I could add that you could not: **both ESP32-P4 targets build.** You flagged that the project-local IDF toolchain was missing on your machine and that the Tanmatsu and T-Display P4 paths were only reviewed statically. I have that toolchain, and both compile. Worth knowing, because that exact gap has bitten three times in the last two days: a symbol defined in `src/main.cpp` broke both P4 targets because neither compiles that file, and twice a declaration inside a board gate vanished on the boards without it.

Which brings me to one thing I fixed on top of your #405, since it is the same class and worth naming so it stops recurring. `s_nav_drawer_corner` and `s_nav_drawer_gear` are declared inside `#if CAP_KEYPAD_NAV` but were assigned unconditionally, in both drawer-close paths and twice while building the grid, so the Heltec V4 TFT and the Wio Tracker L2 did not compile. Guarded to match the declaration. It is invisible on a keypad board, which is exactly why the full matrix is worth running rather than the board in front of you.

Also resolved two conflicts between your own PRs, both by taking the newer decision:

- #412 against #414: the theme chip keeps the sun in both states, which is #414 and matches what DND has done since #396, plus the keypad-nav id from #412. #412 branched before that decision rather than disagreeing with it.
- #409 against #407: the M9 unlock strings keep #407's double-press wording, since #407 changed the gesture itself; your Wio L2 arm is taken as written.

Twelve new user-facing strings went in as translation placeholders so the i18n audit stays clean.

All nine S3 envs and both P4 targets green, your protocol test and the two existing host tests pass.

PR #429, opened 2026-09-06
M9: make special characters selectable

### LINKED ISSUE #410: Still the Problem with special characters on M9 + TANMATSU
Here are two photos showing the problem with the special characters: they appear, but cannot be selected. 

While I'm at it—how do you type other symbols on M9 like $, <, >, %, =, or €? The Tanmatsu has dedicated keys for them, but the M9 doesn't.

Update: TANMATSU is now fine with ALT + ALT GR (see third pic) v0.76

image

image

image

oumike:
## Summary

- add dedicated symbol and emoji controls to M9 channel, direct-message, and room composers
- make the M9 d-pad select picker entries and restore focus to the originating field
- expose the symbol picker in Create/Join channel forms and password fields

## Validation

- VS Code diagnostics: no errors
- `git diff --check`
- PlatformIO build not run; hardware validation remains pending

Fixes #410

### DIFF
--- a/THINKNODE_M9_SHORTCUTS.md
+++ b/THINKNODE_M9_SHORTCUTS.md
@@ -71,6 +71,18 @@ field, so you're never stuck. Enter sends (toggleable under Settings), Back
 leaves the field. In the Terminal, Enter runs the command; in the text editor
 it inserts a newline.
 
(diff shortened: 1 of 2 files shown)


oumike:
I tested this on my M9.

PR #442, opened 2026-09-06
Focus M9 composer when opening chats

### LINKED ISSUE #436: [M9 bug] not possible to start typing when entering chat
There is a problem when entering the chat in which there are no new messages, even though the cursor flashes in the place of typing the text, it is impossible to write and trying to locate where the cursor is now, it finds itself somewhere in chat history

https://github.com/user-attachments/assets/9b7d43bd-b6e0-48eb-a7d0-b41b47ab7c53

oumike:
## Summary
- ignore one-shot focus hints that belong to the screen behind a newly opened chat
- focus the active chat composer when entering a DM, room, or channel
- enter textarea edit mode immediately so M9 keyboard input works without an extra Enter press

## Root cause
The first virtualized-history render could preserve the still-focused inbox row as a valid pointer. Because that object was not part of the newly collected chat focus tree, the hint selected nothing while suppressing the composer fallback. `showKb()` still painted a cursor, leaving visual focus and actual keypad focus out of sync.

## Validation
- hardware tested on ThinkNode M9: opening a chat leaves the composer ready for immediate keyboard input
- `git diff --check` passes
- VS Code diagnostics report no errors in `UITask.cpp`

Fixes #436

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -5077,14 +5077,25 @@ static void navMaybeRebuild() {
   // When a chat just opened, drop focus on its composer so typing goes straight in (issue: textfield
   // not auto-selected). Only on the open transition, so new messages don't steal focus mid-typing.
   bool focus_set = false;   // true once we deliberately focus a COLLECTED obj (else LVGL defaults to the first item)
+  // A still-valid hint can belong to the page behind a newly opened overlay. Only a hint
+  // collected into this rebuild may suppress the new-chat composer fallback (#436).
+  lv_obj_t* collected_focus_hint = nullptr;
   if (s_nav_focus_hint && lv_obj_is_valid(s_nav_focus_hint)) {
     // One-shot explicit focus hint — e.g. turning keyboard-nav ON keeps the highlight on the switch you
     // just toggled instead of snapping the settings list to the top (issue #45).
     const int n = s_nav_count < kNavMax ? s_nav_count : kNavMax;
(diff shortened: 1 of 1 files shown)


B0N3-GD:
Merged, along with #429, #439, #440, #441, #444 and #445. All nine S3 envs and both ESP32-P4 targets build on the merged result.

One conflict worth flagging, because the resolution changes behaviour if it is got wrong. This PR reworks the focus-hint logic so a hint belonging to the page behind a newly opened overlay cannot suppress the new-chat composer fallback. #440 added an M9 pending-focus block immediately above it that places focus deliberately and sets `focus_set`.

Merged naively, the reworked chain fell through to the just-opened-chat branch and moved that focus, which is exactly what #440 exists to prevent. So the chain now has an explicit first arm for the case where focus has already been placed. Both behaviours are kept and the ordering is written down rather than implied.

Separately, and useful for you since you work on Wi-Fi paths: a Heltec V4-R8 crash dump from beta_77 decoded to a data race in the Arduino core, and it is fixed on main. `WiFi.onEvent()` push_backs into a std::vector that the core iterates from its own event task, with no lock on either side. We were registering handlers after `WiFi.mode(WIFI_STA)`, one of them after `WiFi.begin()`, so a STA_DISCONNECTED arriving during boot association could realloc the vector while the event task was walking it, and it then called through a dangling pointer.

Both handlers now register before the stack starts. Worth knowing if you add any more: register them before `WiFi.mode()`, never after, or the window reopens.

PR #469, opened 2026-09-08
Add LilyGo T-Deck Pro support

oumike:
## Summary

- add a PlatformIO board and firmware target for the 3.1-inch LilyGo T-Deck Pro
- support the GDEQ031T10 e-paper display, frontlight, CST328/CST3530 touch, TCA8418 keyboard, SX1262 radio, microSD, GPS, battery reporting, and GPIO0 wake
- coalesce e-paper refreshes with periodic full refreshes and device-specific monochrome UI treatment
- add e-paper map dithering and a monochrome Spectrum snapshot layout
- adapt terminal, conversation focus, scrolling, lock-screen, and keyboard behavior for the Pro
- add T-Deck Pro keyboard-state coverage and port documentation

## Testing

- tested on LilyGo T-Deck Pro hardware
- T-Deck Pro keyboard-state host test passes
- VS Code C++ diagnostics: clean
- `git diff --check`
- branch merged with current upstream `main`

## Additional testing required

**This is a large, cross-cutting new hardware target. Substantial additional testing is required before release.**

Please exercise repeated boot/full-refresh cycles, partial-refresh cadence and ghosting, keyboard input while the panel is BUSY, both CST328 and CST3530 touch revisions, radio TX/RX, SD mount and sustained I/O, GPS, battery reporting, frontlight behavior, GPIO0 sleep/wake, OTA, companion transports, and broad regression coverage across the shared touch UI.


B0N3-GD:
Reviewed. Two findings, one of which needs a change before this can go in.

**The GPS baud probe is correctly inert on every other board, and that is worth saying explicitly** because it is the only substantial ungated change in shared code here. It arms only through the new `HardwareSerial&` constructor overload; every existing board still constructs through the `Stream&` one, so `_probe_serial` stays null and each probe path early-returns on it. That is exactly the right shape for a shared-header addition. The rest of the shared diff is about +723 gated against +139 ungated, and the ungated remainder (the keyboard ring becoming a named power-of-two constant, the e-ink runtime checks, the caps block) is either equivalent or genuinely board-neutral.

**The i18n change is a regression and needs reverting.** This branch takes `{ "System Information", … }` back to `{ "System info", … }` in all thirteen languages, but leaves the source string as `TR("System Information")`. #458 renamed both together; resolving the conflict this way splits them, so the lookup misses and the System Information page title falls back to English in every translated language. Please restore the `"System Information"` keys. The new `"press the button to unlock"` rows in the same file are a good addition and should stay.

On merging: this is a tenth shipping target, and taking it on is @B0N3-GD's call rather than mine, since it pulls in the release matrix, the web flasher, the site and the app-store listing on every future beta. Your own note that it needs substantial hardware testing before release is the other half of that. So I am leaving it for him to decide rather than merging it.

What I can usefully add in the meantime: I am building this branch against all eleven existing targets so the regression question is answered with evidence rather than assumption, and I will post the result here.

B0N3-GD:
Regression result, as promised: this branch builds clean on all nine existing S3 targets, zero errors.

```
heltec_v4_tft            SUCCESS      tlora_pager_lr1121   SUCCESS
heltec_v4_r8_tft         SUCCESS      tlora_pager_sx1262   SUCCESS
attaky_mesh_series       SUCCESS      ThinkNode_M9         SUCCESS
LilyGo_TDeck             SUCCESS      rak_tap_v2           SUCCESS
wio_tracker_l2           SUCCESS
```

Built from a clean worktree at the PR head so nothing local leaked in. That covers the compile-level regression question for the shared changes; it says nothing about runtime behaviour on those boards, which is still worth a pass on real hardware given how much of the touch UI this reaches.

The translation regression I flagged above is still outstanding and is the one thing I would want changed before merge.

oumike:

PR #477, opened 2026-09-09
M9: limit purge confirmation focus

### LINKED ISSUE #432: [M9 bug] skipping thru buttons goes over text in contacts
I don't know if it should work like this but in contacts screen on deleting window you are not only skipping thru buttons but also the text 🤔

https://github.com/user-attachments/assets/9a2b1f50-44f3-4c97-a317-1d11ace06ee1

oumike:
## Summary
- limit keyboard navigation in the “Purge all discovered nodes?” confirmation to Cancel and Purge
- default the dialog selection to Cancel

Fixes #432

## Testing
- VS Code diagnostics: clean
- git diff --check: clean


oumike:

PR #479, opened 2026-09-09
M9: keep composer caret inside text field

### LINKED ISSUE #474: [M9 bug] Cursor jumps out off the text field
"M9 is close to the point where its keyboard navigation stops being something you work around" Yes, almost there. Thanks for all your work!

Unfortunately, there’s still a small cursor issue. When you’re writing a message and use the left arrow key to go back and change or insert a character, you can’t navigate the cursor back to the end of the text. Instead, it jumps out of the text input field.

If you leave the channel without sending the message and then reopen it, the text is still there, but now the cursor is at the end.

Alternatively, instead of moving the cursor left, you can use Backspace to delete the text up to the point where you want to insert something — a little workaround ;-)

oumike:
## Summary

- let Right move the active M9 textarea caret through the final character before leaving edit mode
- preserve the next boundary press as the existing deterministic move to the symbol control

## Validation

- VS Code diagnostics: no errors
- `git diff --check`
- PlatformIO build not run; M9 hardware validation remains pending

Fixes #474

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -41363,6 +41363,15 @@ static bool m9HandleArrowKey(int key, lv_obj_t* ta) {
       return true;
     case M9_KEY_RIGHT:
       {
+        // Reaching the final caret position must not also leave edit mode.
+        if (ta) {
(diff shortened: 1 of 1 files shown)


oumike:

PR #486, opened 2026-09-09
Wio Tracker L2 Tweaks

oumike:
## Summary
- forward normal virtual-keyboard Enter events from the mirror field to the bound dialog field
- dismiss the keyboard safely after room/admin login submission
- use legacy-compatible one-byte path hashes for interactive login floods without changing the saved radio preference
- keep login timeout guidance aligned with the compatibility override

## Validation
- verified Enter submits and closes the keyboard on Wio Tracker L2 hardware
- verified the entered password reaches the login encryption boundary unchanged during diagnosis
- `pio project config --json-output`
- editor diagnostics clean
- `git diff --check`

No PlatformIO build was run after removing temporary diagnostics.


oumike:

PR #493, opened 2026-09-10
Make M9 navigation shortcuts global

### LINKED ISSUE #485: [M9 feature] chats button
Can you guys make that chats button to not only work as a way to enter channels list from Home Screen but also when you are in a channel? So it would work like also like a back button but only to channels list 🤔

oumike:
## Summary
- make the M9 MSG key return an open conversation to the Chats list
- keep MSG, Home, Mentions, Advert, Map, Ctrl, and Advert-hold GPS available through apps, pickers, and active text fields
- preserve contextual Back, d-pad, and OK behavior
- document the global shortcut behavior

Fixes #485

## Testing
- VS Code diagnostics: no errors in touched files
- `git diff --check`

### DIFF
--- a/THINKNODE_M9_SHORTCUTS.md
+++ b/THINKNODE_M9_SHORTCUTS.md
@@ -45,7 +45,7 @@ under a running operation.
 
 | Key | Press | Hold |
 |---|---|---|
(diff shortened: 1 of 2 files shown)


oumike:

PR #494, opened 2026-09-10
Fix M9 admin log scrolling

### LINKED ISSUE #483: [M9 bug] admin page info not scrollable
There is some text on admin page that cannot be seen, no way to scroll that place

https://github.com/user-attachments/assets/1692d35a-20e1-43ec-b583-07f1132b9a0a

oumike:
## Summary
- let the focused M9 admin log consume Up/Down while content remains in that direction
- preserve normal focus movement to the Close header and command row at the log boundaries

Fixes #483

## Testing
- `git diff --check`


oumike:

PR #495, opened 2026-09-10
Show M9 file timestamps directly

### LINKED ISSUE #475: [M9] Crash Report
The file transfer feature is particularly well done — great job! I’ll use it right away to send you a crash report. ;-)

One small feature request for the file browser: Could you display each file’s last-modified date and time from the SD card directly in the file list, ideally next to the filename and file size?

This would be especially useful for crash reports. I have the impression that a new crash report overwrites the previous one rather than creating a new file, so the timestamp would make it immediately clear when the current report was generated.

[wadamesh-crash.elf.txt](https://github.com/user-attachments/files/32013068/wadamesh-crash.elf.txt)

[wadamesh-crash_elf.txt](https://github.com/user-attachments/files/32013106/wadamesh-crash_elf.txt)

oumike:
## Summary
- place M9 file size and modification time on a second line when a valid timestamp exists
- keep long filenames readable while making crash-report dates immediately visible instead of hiding them in the marquee tail
- retain the compact one-line row when the filesystem has no valid timestamp

Fixes #475

## Testing
- verified representative `wadamesh-crash.elf` metadata fits the existing 120-byte row buffer
- `git diff --check`

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -23228,7 +23228,15 @@ static void fmRender() {
         localtime_r(&t, &tmv);
         strftime(ts, sizeof ts, "  %d %b %H:%M", &tmv);
       }
+#if defined(HAS_THINKNODE_M9)
+      // The M9's 320 px row cannot show a crash-report filename, size and date
+      // at once. Keep the filename on line one and make metadata immediately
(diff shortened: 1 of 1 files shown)


oumike:

PR #497, opened 2026-09-11
Add M9 lock-home-to-drawer option

### LINKED ISSUE #491: Home button behavior
Is it possible to set that home button like: when you have app drawer selected as home then it will not switch between app drawer and that cmdr home? So it will always stay on app drawer home and if you need to enter that "other" Home Screen then you can choose it from app drawer 🤔

https://github.com/user-attachments/assets/81e527fd-2444-486d-a60f-aad3399fd3ed

oumike:
## Summary
- add an M9-only **Lock home to drawer** option in the app-drawer cog settings
- show the option only while **App drawer as home** is enabled
- keep HOME in or return it to the drawer while preserving the live drawer position; the Cmdr tile still opens Commander
- persist the option with an append-only v60 preference migration that defaults off

Fixes #491

## Testing
- `c++ -std=c++17 -Isrc test/test_touch_prefs_schema.cpp -o /tmp/wadamesh-test-touch-prefs-schema && /tmp/wadamesh-test-touch-prefs-schema`
- VS Code diagnostics: no errors in touched files
- `git diff --check`

### DIFF
--- a/THINKNODE_M9_SHORTCUTS.md
+++ b/THINKNODE_M9_SHORTCUTS.md
@@ -36,6 +36,9 @@ because that is the root.
 
 **HOME** is the shortcut past all of that: it goes straight to the root and
 clears the trail, so Back from there won't jump you back out again.
+When **App drawer as home** is enabled, the drawer cog's **Lock home to drawer**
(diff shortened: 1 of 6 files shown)


oumike:

PR #498, opened 2026-09-11
Fix battery estimate after clock changes

### LINKED ISSUE #480: [bug] Battery life showing charging / steady instead of value since Beta 79
Since Beta 79 I now no longer see the battery life value showing on the M9. I did a full restart again to be sure.

Image

oumike:
## Summary
- ignore invalid battery-log epochs when estimating remaining life
- restart the regression at the newest backward or duplicate timestamp after a reboot or clock correction
- preserve the existing charge-event segmentation and 24-hour chart history

Fixes #480

## Testing
- reproduced a descending battery history crossing a backward clock jump: the old calculation produced a near-zero slope, while the fixed calculation used the new monotonic segment and returned a finite estimate
- `git diff --check`


oumike:

PR #501, opened 2026-09-11
Fix Wio Tracker L2 backups on SD

### LINKED ISSUE #499: [bug][Wio Tracker L2] "Export new backup" never writes to the SD card — the SD path is compiled out (CAP_SD 0)
## Summary

On the **Seeed Wio Tracker L2** build (`wio_tracker_l2_companion_radio_touch`), Settings → Backups → **Export new backup** always lands on internal flash (SPIFFS), even with a working microSD inserted. The toast reads `Saved internal:/meshcore-….json`, never `Saved SD:/…`. The same applies to Profile → "Export settings" (`meshcore-backup.json`), to the backup **scan** (SD-resident `.json` files never appear in the picker), and to **delete**.

## Board / build

- Board: Seeed Wio Tracker L2 (ESP32-S3), env `wio_tracker_l2_companion_radio_touch`
- Firmware: current `main` (beta_76 line)
- Card: mounts fine — it's adopted as the data store at boot (`[BOOT] storage: SD /meshcomod`), the file manager browses it, and screenshots save to it

## Expected vs actual

- **Expected:** with a card present, the export lands on the SD card and is listed as an `SD:` entry in the Backups list — i.e. removable, matching the button's documented behaviour ("goes to the SD card if one is in, else internal flash", `src/ui-touch/UITask.cpp:39140`).
- **Actual:** the export always goes to internal SPIFFS. A user who pulls the card to copy the backup off the device finds nothing there.

## Root cause

Every SD branch in the backups code is gated on `#if CAP_SD || defined(TLORA_PAGER)`, and on this board `CAP_SD` is **0**, because its card is on **SD_MMC**, not the Arduino `SD` library:

```c
// src/ui-touch/device_caps.h:28-32 — Seeed Wio Tracker L2
#define CAP_SD           0   // SD_MMC, not Arduino SD
```

So the preprocessor removes the card entirely from the backups feature on this target:

- `doExportBackupFile()` — `src/ui-touch/UITask.cpp:38993`, SD branch at `:39010`. With the branch gone, `f` is always opened on `backupInternalFs()` (SPIFFS here) and `where` stays `"internal"`.
- `backupScan()` — `src/ui-touch/UITask.cpp:38800`. SD root and `/meshcomod` are never scanned, so a `meshcore-backup.json` copied onto the card is invisible to the import picker.
- `doDeleteBackup()` — `src/ui-touch/UITask.cpp:39032`, plus the other backup-adjacent gates at `:39048` and `:39092`.

Note the gate isn't just about a missing card at runtime — it's compile-time, so the code never even looks.

## Why the card is otherwise fine on this board

The Wio Tracker L2 has a fully working mount path that other features already use:

- boot mount: `src/main.cpp:1020` (`SD_MMC.setPins(2, 3, 1)`, 1-bit, `SDMMC_FREQ_DEFAULT`)
- UI-side remount helper compiled for this board: `src/ui-touch/UITask.cpp:22509` / `:22525`
- file manager treats `&SD_MMC` as the SD root: `src/ui-touch/UITask.cpp:19813`, `:22566`

## Suggested fix

Follow the pattern `takeScreenshotToSd()` already uses at `src/ui-touch/UITask.cpp:43789-43793`: widen the gate and select the filesystem object per board, rather than assuming `SD`.

```c
#if CAP_SD || defined(TLORA_PAGER) || defined(HAS_WIO_TRACKER_L2)
  #if defined(HAS_WIO_TRACKER_L2)
    if (tanSdTryMount()) { f = SD_MMC.open(path, FILE_WRITE); if (f) where = "SD"; }
  #else
    if (fmSdTryMount())  { f = SD.open(path, FILE_WRITE);     if (f) where = "SD"; }
  #endif
#endif
```

The same `fs::FS*` selection is needed in `backupScan()` (scan `/` and `/meshcomod`) and `doDeleteBackup()` so exported files round-trip: list → import → delete.

Two things worth deciding while fixing:

1. The other SD_MMC boards (Tanmatsu `CAP_SD 0`, T-Display P4, RAK Tap V2) are in the same position — a shared `backupSdFs()` helper returning `fs::FS*` (or `nullptr` when no card) would fix the family in one place instead of adding a third `#if` flavour to each of the ~8 backup call sites.
2. On this board the SD is the *data store* (`/meshcomod`), so it's worth choosing deliberately whether the export lands at the card root (most discoverable when the card goes into a PC) or next to the rest of the data — `backupScan()` already looks in both, so the root seems right.

## Related

- #231 — "[bug][T-Display P4][beta 56] backup can't open file": the same feature breaking on another non-Arduino-`SD` board, fixed there by correcting `backupInternalFs()`. This is the remaining half of that story — the SD side.

---
_Generated by [Claude Code](https://claude.ai/code)_

oumike:
## Summary
- add a shared backup filesystem selector for SPI SD and SD_MMC boards
- use the selected filesystem for backup export, scan, import, and delete
- preserve the internal-storage fallback when no card is available

## Testing
- `platformio run --target upload --environment wio_tracker_l2_companion_radio_touch`
- tested on a Seeed Wio Tracker L2

Fixes #499

### DIFF
--- a/src/ui-touch/UITask.cpp
+++ b/src/ui-touch/UITask.cpp
@@ -40391,6 +40391,15 @@ static fs::FS* backupInternalFs() {
   return &SPIFFS;
 #endif
 }
+static fs::FS* backupSdFs() {
+#if CAP_SD || defined(TLORA_PAGER)
+  return fmSdTryMount() ? &SD : nullptr;
+#elif defined(HAS_TANMATSU) || defined(HAS_TDISPLAY_P4) || defined(HAS_WIO_TRACKER_L2)
+  return tanSdTryMount() ? &SD_MMC : nullptr;
+#else
+  return nullptr;
+#endif
+}
 static void backupScan() {
   s_backup_count = 0;
   // Internal flash is flat — list any *.json at the root (SPIFFS on T-Deck/V4, FFat on Tanmatsu).
@@ -40410,17 +40419,15 @@ static void backupScan() {
     }
     root.close();
   }
-#if CAP_SD || defined(TLORA_PAGER)
   // SD card: scan the root AND the /meshcomod data folder. SD-storage builds keep
   // their data under /meshcomod, so users naturally drop a backup next to it —
-  // previously only the root was scanned, so a json in /meshcomod never showed up
-  // ("only gets recognized from the root folder"). Actively (re)mount here: the SD
-  // is otherwise only probed while the file manager is open. fmSdTryMount() no-ops
-  // if already mounted, else walks the mount ladder.
-  if (fmSdTryMount()) {
+  // previously only the root was scanned, so a json in /meshcomod never showed up.
+  // backupSdFs() adopts or mounts either the board's SPI SD or SD_MMC card.
+  fs::FS* sd_fs = backupSdFs();
+  if (sd_fs) {
     static const char* kSdDirs[] = { "/", "/meshcomod" };
     for (const char* dir : kSdDirs) {
-      File d = SD.open(dir);
+      File d = sd_fs->open(dir);
       if (d && d.isDirectory()) {
         File e = d.openNextFile();
         while (e) {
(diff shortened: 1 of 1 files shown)


oumike:

PR #502, opened 2026-09-11
Allow longer quick replies

### LINKED ISSUE #425: [feature] Quick replies with longer text allowed
The quick replies actually work very well for most use cases like we use it for "Happy #MeshCoreMonday Cape Town" every week and that fits in fine.

I also advertise a weekly radio bulletin out on the MeshCore mesh and the message is a bit longer as it includes the link to listen at. It is short enough to fit into a single MeshCore message post, but it seems the WADAMESH quick replies is only allowing quite short messages.

Could it be made longer? I don't think it will have any detrimental affect anywhere as short messages will still work as before.  I'm looking for at least 65 characters which would work, but maybe 100 is more flexible for others?

oumike:
## Summary
- increase configurable quick replies from 31 to 100 characters
- keep editor input, persistence truncation, picker display, and composer insertion on the shared limit
- preserve existing quick-reply keys, defaults, and shorter saved values

## Testing
- `platformio run --target upload --environment wio_tracker_l2_companion_radio_touch`
- tested on a Seeed Wio Tracker L2

Fixes #425

### DIFF
--- a/src/helpers/esp32/TouchPrefsStore.h
+++ b/src/helpers/esp32/TouchPrefsStore.h
@@ -89,11 +89,11 @@ bool    touchPrefsSetLangFile(const char* code);
 
 /** User-configurable quick-reply macros: up to 6 short strings the user can
  *  drop into the composer with a single tap (e.g. "ok", "on the way",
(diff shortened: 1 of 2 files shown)


oumike:

PR #513, opened 2026-09-13
Fix Lua map lifetime across ui.clear()

### LINKED ISSUE #511: `ui.clear()` does not invalidate a live `wada.map` view — stale handle, stuck view count, and `lv_obj_del` on a deleted object
Reading the Lua host while writing a two-screen app, `MapUd` looks like the one handle type that does not
participate in the generation scheme, and the consequences land exactly on the screen-switch idiom the SDK
recommends. I have not run this — it is from the source at `16d0969` — so please sanity-check me, but the
paths look unambiguous.

### Every other handle self-invalidates; the map handle has no generation

```
src/ui-touch/LuaAppHost.cpp:337   struct CanvasUd { lv_obj_t* obj; lv_color_t* buf; int w, h; uint32_t gen; };
src/ui-touch/LuaAppHost.cpp:446   struct WidgetUd { lv_obj_t* obj; uint32_t gen; };
src/ui-touch/LuaAppHost.cpp:509   struct ChartUd  { lv_obj_t* obj; lv_chart_series_t* ser[2]; int n_ser; uint32_t gen; };
src/ui-touch/LuaAppHost.cpp:647   struct ListUd   { lv_obj_t* obj; int sel; uint32_t gen; };
src/ui-touch/LuaAppHost.cpp:1392  struct MapUd    { void* v; };          <-- no gen
```

and each of the others is checked on access, e.g.

```
src/ui-touch/LuaAppHost.cpp:449   if (u->gen != s_ui_gen) u->obj = nullptr;   // cleared out from under this handle
src/ui-touch/LuaAppHost.cpp:512   if (u->gen != s_ui_gen) { u->obj = nullptr; u->n_ser = 0; }
src/ui-touch/LuaAppHost.cpp:650   if (u->gen != s_ui_gen) u->obj = nullptr;
```

`checkMap` (`:1394`) has no equivalent. The comment above `uiClear` explains precisely why the others need
it — *"A Lua variable still holding a label from the screen just cleared keeps a raw pointer to freed
memory; the accessors compare generations and null it"* (`:711-714`) — and the map is the one handle that
does not get that treatment.

### Why that bites

The view's container is a **child of the app body**:

```
src/ui-touch/UITask.cpp:31044      v->cont = lv_obj_create(body);       // body = luaHostAppBody()
```

and `ui.clear()` deletes exactly that body's children:

```
src/ui-touch/LuaAppHost.cpp:718    lv_obj_clean(s_h->body);   // deletes the children, keeps the page itself
src/ui-touch/LuaAppHost.cpp:719    ++s_ui_gen;
```

So after `ui.clear()` with a view still open, three things are true at once:

1. **`v->cont` dangles.** Whenever `:close()` or `__gc` eventually runs, `luaHostMapDestroy` reaches
   `if (v->cont) lv_obj_del(v->cont);` (`UITask.cpp:31069`) on an object LVGL has already freed.
   `mpGc` is `mpClose` (`LuaAppHost.cpp:1481`), so collection alone is enough to get there.
2. **`s_map_views` stays 1**, because only `mpClose` decrements it (`:1479`). The next
   `wada.map.view()` therefore fails with `only one map view at a time (call :close() first)` (`:1399`) —
   and the app has no safe way out, since calling `:close()` on the stale handle is what triggers (1).
3. **The tile pool is stranded** until that unsafe close happens: `appMapFreeTile` over
   `kAppMapTilesMax` only runs inside `luaHostMapDestroy` (`UITask.cpp:31068`), and the doc puts that at
   four decoded tiles ≈ 512 KB (`sdk.txt:169`).

The app-teardown path already anticipates this shape and guards it deliberately:

```
src/ui-touch/LuaAppHost.cpp:2704-2707
  // A map view is torn down with the app body below, so its userdata __gc must
  // not free it a second time. Zeroing the count here also lets the next app
  // open a view even if the previous one never called :close().
  s_map_views = 0;
```

`uiClear` never got the same guard, which is what makes this look like an oversight rather than a design
choice.

### Why it matters more than it looks

`ui.clear()` is the documented way to build a multi-screen app — *"This is how you build an app with more
than one screen: clear, then create the next screen's widgets in the same place you built the first"*
(`sdk.txt:82`). So any app that puts a map on one screen and something else on another walks into this
unless its author happens to close the view **before** clearing. The ordering is invisible in the docs and
there is nothing that warns you.

### Reproduction (untested — this is the shape, not an observed run)

```lua
local app, map = {}, nil
function app.on_open(w, h)
  map = wada.map.view(4, 4, w - 8, 100)
  wada.ui.button("swap", 4, 110, 80, 32, function()
    wada.ui.clear()                       -- deletes the view's container
    local ok, err = pcall(function() return wada.map.view(4, 4, w - 8, 100) end)
    wada.sys.toast(ok and "second view ok" or tostring(err))   -- expect the one-view error
    map:close()                           -- expect lv_obj_del on freed memory
  end)
end
return app
```

### Possible direction

Since `s_map_views` is capped at one, the host could hold the live view pointer beside the count and have
`uiClear` call `luaHostMapDestroy` on it before `lv_obj_clean`, then zero the count — which frees the tile
pool properly and leaves nothing dangling, mirroring what the app-close path already intends. Giving
`MapUd` a `gen` and checking it in `checkMap` would stop the stale `:close()`, but on its own it would
leak the `AppMapView` and its tiles, so the two probably want to go together. Entirely your call — I have
flagged it rather than assumed a fix.

Happy to test a patch if it would help; I have a T-Deck Plus.

oumike:
## Summary
- track the active Lua map view directly instead of maintaining a separate count
- destroy the live map and free its tile pool before `ui.clear()` removes body children
- generation-invalidate stale map userdata so later `:close()` and `__gc` calls are harmless
- explicitly release any live map before Lua finalizers run during app teardown

## Validation
- VS Code C++ diagnostics: no errors
- `git diff --check`
- PlatformIO build not run

Fixes #511


oumike:

PR #514, opened 2026-09-13
Add universal microSD diagnostics

oumike:
## Summary
- add a physical microSD capability independent of the SPI `SD` versus `SD_MMC` backend
- record existing boot, lazy-mount, and recovery attempts without adding new mount probes
- show mounted state, attempt count, initialization result, and operating clock under Settings > About > System Information
- cover T-Deck, T-LoRa Pager, ThinkNode M9, Heltec V4-R8, Wio Tracker L2, Tanmatsu, and T-Display P4
- report that microSD is unavailable on hardware without a card slot

## Validation
- `platformio run --target upload --environment ThinkNode_M9_companion_radio_touch`
- `platformio run --target upload --environment LilyGo_TDeck_companion_radio_touch`
- VS Code C++ diagnostics: no errors in all changed files
- `git diff --check`

SD_MMC diagnostics remain passive: opening System Information does not add card traffic on those targets.

PR #509, opened 2026-09-13
Fix M9 new-channel composer focus

oumike:
## Summary
- open a newly created private channel immediately on M9
- restore composer focus after the empty-chat placeholder rebuild
- keep the symbol button reachable from the composer with directional navigation

## Testing
- Tested on ThinkNode M9 hardware

Fixes #508

PR #492, opened 2026-09-10
Improve M9 chat navigation

oumike:
## Summary
- show the existing jump-to-latest arrow on M9 when newer messages are below the viewport
- let Right from a focused message select the arrow so OK jumps to the newest message
- hide bottom mail/contact notification indicators throughout the Chats tab to prevent redraw overlap
- document the M9 shortcut

Fixes #490

## Testing
- Tested on ThinkNode M9 hardware
- VS Code diagnostics: no errors in touched files
- `git diff --check`

PR #489, opened 2026-09-09
Fix At-a-glance zero-hop count

oumike:
## Summary
- decode the packed flood-path descriptor before formatting the At-a-glance hop count
- preserve unknown-path handling and direct/singular/plural labels

Fixes #484

## Testing
- verified packed values: `0x40 -> 0`, `0x41 -> 1`, `0x82 -> 2` hops
- VS Code diagnostics: no errors in `src/ui-touch/UITask.cpp`
- `git diff --check`

PR #517, opened 2026-09-15
Consolidate all active development changes

### LINKED ISSUE #425: [feature] Quick replies with longer text allowed

### LINKED ISSUE #427: [M9 bug] cannot see all stats in about
We are not able to view all stats in about screen since there scroll is jumping thru buttons 

https://github.com/user-attachments/assets/1caaf5f7-cda6-41d1-b649-4234d2b17564

### LINKED ISSUE #432: [M9 bug] skipping thru buttons goes over text in contacts

### LINKED ISSUE #434: [M9 bug] highlights colors
I think that M9 highlights are acting strange sometimes, like in control menu on the brightness it's with the color that you picked in display settings (?), but when you move to power button it's ugly not even round white 😉

https://github.com/user-attachments/assets/54bce883-231a-44e7-be06-0862aa3b5947

Same with duty meter, it's so ugly white 😁

https://github.com/user-attachments/assets/e473a754-9fdf-4daa-8b9b-3c2f4124f177

Why can't be highlights everywhere like in app drawer menu? Somehow related to chosen display color

https://github.com/user-attachments/assets/4ff31920-ba95-4e24-8ff9-6acdbbb6ca8e

In chats there is even blue color somehow not related to chosen one, on tdeck this is a different one:

image

https://github.com/user-attachments/assets/00c0a407-c196-4f23-978d-4047b372029c

### LINKED ISSUE #474: [M9 bug] Cursor jumps out off the text field

### LINKED ISSUE #475: [M9] Crash Report

### LINKED ISSUE #480: [bug] Battery life showing charging / steady instead of value since Beta 79

### LINKED ISSUE #481: [M9 bug] discovered contacts scrolling
There is an issue with discovered contacts scrolling up. It's jumping between add and close

https://github.com/user-attachments/assets/d0d3cfd4-1a7c-46b3-9222-55b3fdba5bfb

### LINKED ISSUE #482: [M9 bug] cannot scroll message info up
There is no way to scroll message info up when there are many hops

https://github.com/user-attachments/assets/1b7570a2-c58f-44b7-b9e9-462d191a2ea6

### LINKED ISSUE #483: [M9 bug] admin page info not scrollable

### LINKED ISSUE #484: At a glance 0 hops showed as 64
Like in the topic, at a glance shows 0 hops as 64 😉

image

### LINKED ISSUE #485: [M9 feature] chats button

### LINKED ISSUE #490: Missing go to bottom arrow in channel chats?
Hi, I think the down arrow is missing somewhere in the chats, now when you have a lot of new messages you have to press the down button many times or exit and enter the channel again. This arrow would be a great option to jump to the very bottom, for example, being on some message and clicking on the right could go to it. Unless there is an option to scroll by holding the down button? 🤔

https://github.com/user-attachments/assets/2dad9d18-e9b7-49ab-a707-f1db817b476d

### LINKED ISSUE #491: Home button behavior

### LINKED ISSUE #499: [bug][Wio Tracker L2] "Export new backup" never writes to the SD card — the SD path is compiled out (CAP_SD 0)

### LINKED ISSUE #508: [M9] New Channel
When you create a new channel, you can not write a first own message until you have received a message from someone else. The cursor does not jump to the input field.

### LINKED ISSUE #511: `ui.clear()` does not invalidate a live `wada.map` view — stale handle, stuck view count, and `lv_obj_del` on a deleted object

oumike:
## Summary

This PR consolidates every pull request that was open against `main` on 15 September 2026. The original commits and authorship are preserved in the merge history, including contributions from `aigarslv` and `tmetz1987`.

It combines:
- LilyGo T-Deck Pro support
- M9 navigation, focus, scrolling, file-layout, and lock behavior fixes
- Wio Tracker L2 behavior and SD_MMC backup fixes
- battery clock-discontinuity and at-a-glance hop-count fixes
- longer quick replies and Latvian keyboard support
- Wardrive 1.2 on-device review
- Lua map lifetime safety across `ui.clear()`
- universal microSD diagnostics
- aggregate Pager and ESP-IDF build compatibility fixes

## Supersedes

- #469 Add LilyGo T-Deck Pro support
- #476 M9 settings-detail navigation
- #477 M9 purge confirmation focus
- #478 M9 navigation cursor highlights
- #479 M9 composer caret behavior
- #486 Wio Tracker L2 tweaks
- #487 M9 discovered-contact scrolling
- #488 M9 message-info scrolling
- #489 At-a-glance zero-hop count
- #492 M9 chat navigation
- #493 M9 global navigation shortcuts
- #494 M9 admin-log scrolling
- #495 M9 file timestamp layout
- #496 Latvian keyboard layout
- #497 M9 lock-home-to-drawer option
- #498 Battery estimate after clock changes
- #501 Wio Tracker L2 SD_MMC backups
- #502 Longer quick replies
- #509 M9 new-channel composer focus
- #510 Wardrive 1.2 review screen
- #513 Lua map lifetime across `ui.clear()`
- #514 Universal microSD diagnostics

## Integration notes

- Composed the M9 Right-key behavior from #479 and #492 so textarea caret movement and latest-message navigation both remain available.
- Composed the independent scrolling behavior from #488 and #494 so message-info and admin-log panes both consume vertical navigation correctly.
- Added the fixed 4 MHz Pager SD clock to the universal diagnostics path.
- Made Tanmatsu and T-Display P4 builds reproducible under the project-local ESP-IDF environment, including paths containing spaces.
- Regenerated the baked Lua source after merging Wardrive 1.2.

## Validation

PlatformIO targets verified on the aggregate tree:
- Heltec V4 TFT
- Heltec V4-R8 TFT
- Attaky Mesh Series
- LilyGo T-Deck
- LilyGo T-Deck Pro
- Wio Tracker L2
- T-LoRa Pager LR1121
- T-LoRa Pager SX1262
- ThinkNode M9
- RAK WisMesh Tap V2

ESP-IDF matrix:
- Tanmatsu: OK
- T-Display P4: OK

Additional checks:
- VS Code C++ diagnostics: no errors in primary merged files
- shell syntax checks for build wrappers
- `git diff --check`
- all 22 superseded PR heads verified as ancestors of this branch
## Linked issues

Fixes #425
Fixes #427
Fixes #432
Fixes #434
Fixes #474
Fixes #475
Fixes #480
Fixes #481
Fixes #482
Fixes #483
Fixes #484
Fixes #485
Fixes #490
Fixes #491
Fixes #499
Fixes #508
Fixes #511


B0N3-GD:
Merged into `main` as 929356d, with one piece held back: the MQTT Observer.

@oumike, thank you for consolidating this, it made the review tractable. Everything up to 703afb9 is in, with your commits and those of @aigarslv and @tmetz1987 kept exactly as they were (same SHAs, same authorship). The only thing left out is the final merge that brought in the Observer for #518 (618ffd0 via c4bfc93). That is a product decision rather than a code-quality one, so it now has its own draft PR, #520, where the review notes are written up and the discussion can happen without holding the rest of this back. Nothing in the earlier history depended on it.

Because `main` took 703afb9 rather than this PR's head, GitHub will not show this PR as merged, hence closing it by hand.

**One fix on top.** The "System Information" translation key was reverting in all 13 languages. That was not this PR's doing: `i18n_builtin.h` is generated from `deploy/apps/lang/*.lang`, and #458's rename changed the generated file but never the packs, so every regeneration, including the one that produced your copy, wrote the old key back. Fixed at the source in 6bb98a1.

**Verified before pushing:** the 10 PlatformIO targets including the T-Deck Pro, both T-Display P4 SKUs and the Tanmatsu, plus the host tests (11 pass, including the new T-Deck Pro keyboard test).

**On the T-Deck Pro:** it now builds on `main` but is not yet in `release.sh`, the flasher metadata or the site, so it will not ship in a beta until @B0N3-GD decides to publish it. That matches your own note that it needs substantial hardware testing first.

**Linked issues:** I am leaving them open until the reporters confirm on the next beta rather than closing them now.

